Issue #136, opened 2020-02-26
Problem with reading


mano-professional:
I am able to establish a connection but there is some problem with reading the columns. When I do df.columns, I see an empty list. When I do df.info_es(), I see the following error.

`AttributeError                            Traceback (most recent call last)
<ipython-input-10-3038caf276b4> in <module>
----> 1 df.info_es()

~/.local/lib/python3.7/site-packages/eland/dataframe.py in info_es(self)
    548         buf = StringIO()
    549 
--> 550         super()._info_es(buf)
    551 
    552         return buf.getvalue()

~/.local/lib/python3.7/site-packages/eland/ndframe.py in _info_es(self, buf)
    149 
    150     def _info_es(self, buf):
--> 151         self._query_compiler.info_es(buf)
    152 
    153     def mean(self, numeric_only=True):

~/.local/lib/python3.7/site-packages/eland/query_compiler.py in info_es(self, buf)
    471         self._index.info_es(buf)
    472         self._mappings.info_es(buf)
--> 473         self._operations.info_es(self, buf)
    474 
    475     def describe(self):

~/.local/lib/python3.7/site-packages/eland/operations.py in info_es(self, query_compiler, buf)
    774         query_params, post_processing = self._resolve_tasks(query_compiler)
    775         size, sort_params = Operations._query_params_to_size_and_sort(query_params)
--> 776         _source = query_compiler._mappings.get_field_names()
    777 
    778         script_fields = query_params['query_script_fields']

~/.local/lib/python3.7/site-packages/eland/field_mappings.py in get_field_names(self, include_scripted_fields)
    548     def get_field_names(self, include_scripted_fields=True):
    549         if include_scripted_fields:
--> 550             return self._mappings_capabilities.es_field_name.to_list()
    551 
    552         return self._mappings_capabilities[

~/.local/lib/python3.7/site-packages/pandas/core/generic.py in __getattr__(self, name)
   5177             if self._info_axis._can_hold_identifiers_and_holds_name(name):
   5178                 return self[name]
-> 5179             return object.__getattribute__(self, name)
   5180 
   5181     def __setattr__(self, name, value):

AttributeError: 'DataFrame' object has no attribute 'es_field_name'`

stevedodson:
@mano-professional - many thanks for the feedback. Can you share some details about the index mapping and number of documents in the index? 

Also, https://github.com/elastic/eland/pull/135 fixed an issue in this space, and release this fix shortly.

[...]

PR #149, opened 2020-03-24
Bug/is scripted error

stevedodson:
Minor mod to support 6.x style indices.

Currently, there is no specific test for this as
it requires a 6.x cluster. 6.x is not officially
supported by 7.x clients, but this is a friendly
option for users.

Longer term we should add a version check at the
high level.


stevedodson:
Hopefully resolves https://github.com/elastic/eland/issues/136

eland/field_mappings.py
```diff
@@ -162,8 +183,16 @@ def flatten(x, name=''):
         for index in mappings:
             if 'properties' in mappings[index]['mappings']:
                 properties = mappings[index]['mappings']['properties']
-
-                flatten(properties)
+            else:
```
sethmlarson: Could we add a unit test so we don't regress this behavior?

eland/field_mappings.py
```diff
@@ -162,8 +183,16 @@ def flatten(x, name=''):
...
+                # Pre Elasticsearch 7.0 mappings had types. Support these
+                # in case eland is connected to 6.x index - this is not
+                # officially supported, but does help usability
+                es_types = list(mappings[index]['mappings'].keys())
+                if len(es_types) != 1:
+                    raise NotImplementedError("eland only supports 0 or 1 Elasticsearch types")
```
sethmlarson: Maybe this exception can be augmented to add the `es_types` field to give the user an indication what we're seeing on our end?

stevedodson:
@sethmlarson - updates based on your comments.

Whilst I'm not a big fan of this change (I like strictness), I have found it very useful in practise when accessing 6.x clusters. 

We should probably add a high level check to warn if you connect to a 6.x cluster, we don't fully support or fully test.

Note, strictly the elasticsearch-py client shouldn't support 6.x clusters.

stevedodson:
Fails CI on python 3.5 as I changed `OrderedDict` to `dict`.

sethmlarson:
The f-string might cause an issue too, glad to see you're already in 3.6 mode :)

PR #153, opened 2020-03-28
Issue/140 show version number

### LINKED ISSUE #140: Call to eland.__version__ should show version number
If I'm working in the Python shell with pandas and want to know which version I have installed, I can do a handy call to 

```
import pandas as pd
pd.__version__
```
It would be nice if `eland` supported a similar functionality.
At the moment, the behaviour is as follows.

```
Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas as pd
>>> pd.__version__
'0.25.3'
>>> import eland as ed
>>> ed.__version__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'eland' has no attribute '__version__'
>>> 
```

mesejo:
* Show the version number
    
   The version number was being imported using the following statement:
    `from eland._version import *`
    This way only loads the non-private names (see [this](https://stackoverflow.com/a/1361852/4001592)), importing the names implicitly solves the issue. 

*  Add ELLIPSIS, to ignore trailing white-spaces in doctest.

Closes #140


stevedodson:
jenkins test this please

eland/__init__.py
```diff
@@ -14,7 +14,16 @@
 
 # flake8: noqa
 
-from eland._version import *
+from eland._version import (
+    __title__,
+    __description__,
```
Winterflower: Looking over this change, I actually much prefer it when the imports are listed out like in the lines above instead of just using the asterisk - it makes it easier to find where things are defined (imho).  @sethmlarson is it common practice to do asterisk imports in __init__.py? I guess it saves some typing, but I've always generally had the feel that asterisk imports in general were frowned upon.

sethmlarson: Yeah star imports are not optimal, makes it hard for tooling to figure out the true signature of objects being used without searching everywhere. I actually have a flake warning suppressed due to this but I'm saving the fix until we narrow our public API so there's no rework to do. :)

PR #154, opened 2020-03-31
Adding ignore_index to pandas_to_eland

stevedodson:
Parameter is useful when adding multiple pd.DataFrame's to
the same index.

Also, updated test module to pandas.testing for 1.0.x
compliance,

### DIFF
--- a/eland/tests/dataframe/test_utils_pytest.py
+++ b/eland/tests/dataframe/test_utils_pytest.py
@@ -65,6 +65,40 @@ class TestDataFrameUtils(TestData):
 
         assert_pandas_eland_frame_equal(df, ed_df_head)
 
+        ES_TEST_CLIENT.indices.delete(index=index_name)
+
+    def test_pandas_to_eland_ignore_index(self):
+        df = pd.DataFrame(
+            data={
+                "A": np.random.rand(3),
+                "B": 1,
+                "C": "foo",
+                "D": pd.Timestamp("20190102"),
+                "E": [1.0, 2.0, 3.0],
+                "F": False,
+                "G": [1, 2, 3],
+            },
+            index=["0", "1", "2"],
+        )
+
+        # Now create index
+        index_name = "test_pandas_to_eland_ignore_index"
+
+        ed_df = ed.pandas_to_eland(
+            df,
+            ES_TEST_CLIENT,
+            index_name,
+            es_if_exists="replace",
+            es_refresh=True,
+            ignore_index=True,
+        )
+        pd_df = ed.eland_to_pandas(ed_df)
+
+        # Compare values excluding index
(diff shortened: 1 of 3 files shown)


eland/utils.py
```diff
@@ -185,16 +189,19 @@ def pandas_to_eland(
...
-        id = row[0]
-
         if es_dropna:
             values = row[1].dropna().to_dict()
         else:
             values = row[1].to_dict()
 
-        # Use integer as id field for repeatable results
-        action = {"_index": es_dest_index, "_source": values, "_id": str(id)}
+        if ignore_index:
```
sethmlarson: Maybe we should change the name of this parameter to be more verbose but also clear on what "index" is being ignored since the concept appears in both pandas and ES.

Maybe `ignore_pandas_index` or even flip the default and use `use_pandas_index_as_id=True`?

eland/utils.py
```diff
@@ -86,7 +87,10 @@ def pandas_to_eland(
     es_geo_points: list, default None
         List of columns to map to geo_point data type
     chunksize: int, default None
-        number of pandas.DataFrame rows to read before bulk index into Elasticsearch
+        Number of pandas.DataFrame rows to read before bulk index into Elasticsearch
+    ignore_index: bool, default 'False'
+        Ignore pandas.DataFrame.index when indexing into Elasticsearch?
```
sethmlarson: Drop the ?, document the behavior for the true and false case

eland/tests/dataframe/test_utils_pytest.py
```diff
@@ -65,6 +65,40 @@ def test_generate_es_mappings(self):
 
         assert_pandas_eland_frame_equal(df, ed_df_head)
 
+        ES_TEST_CLIENT.indices.delete(index=index_name)
+
+    def test_pandas_to_eland_ignore_index(self):
```
sethmlarson: This unit test passes without the `ignore_index=True`, should add a that fails if not using the functionality.

eland/tests/dataframe/test_utils_pytest.py
```diff
@@ -65,6 +65,40 @@ def test_generate_es_mappings(self):
...
+            ES_TEST_CLIENT,
+            index_name,
+            es_if_exists="replace",
+            es_refresh=True,
+            ignore_index=True,
+        )
+        pd_df = ed.eland_to_pandas(ed_df)
+
+        # Compare values excluding index
+        assert df.values.all() == pd_df.values.all()
```
sethmlarson: Shouldn't this be `(df.values == pd_df.values).all()`?
Also don't we want to compare index of these frame to show that they are not the same?

stevedodson:
Many thanks! Great comments - resolved in new commit.

PR #166, opened 2020-04-03
Remove eland.Client, use Elasticsearch directly

sethmlarson:
Now that Elasticsearch has up-to-date APIs for 7.6+ we shouldn't need a shim with direct access to `perform_request()`. If we do go back to having a shim between `QueryCompiler` and `elasticsearch-py` it probably won't be a public interface.


sethmlarson:
@stevedodson I have a plan to use [MonkeyType](https://github.com/Instagram/MonkeyType) and run the test suite to get a really good starting point then amend the parts that are missing, we'll see how well it works with pandas data types. :)

PR #176, opened 2020-04-13
Add median()  to ndframe

mesejo:
Includes:
- Add implementation of median
- Add docstring of std, var and, median
- Run black

### DIFF
--- a/eland/ndframe.py
+++ b/eland/ndframe.py
@@ -267,7 +267,7 @@ class NDFrame(ABC):
         Examples
         --------
         >>> df = ed.DataFrame('localhost', 'flights')
-        >>> df.std() # doctest: +SKIP
+        >>> df.var() # doctest: +SKIP
         AvgTicketPrice        7.096185e+04
         Cancelled             1.119831e-01
         DistanceKilometers    2.096049e+07
--- a/eland/ndframe.py
+++ b/eland/ndframe.py
@@ -252,10 +252,94 @@ class NDFrame(ABC):
         return self._query_compiler.min(numeric_only=numeric_only)
 
     def var(self, numeric_only=True):
+        """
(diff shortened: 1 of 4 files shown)


sethmlarson:
Jenkins test this please

eland/ndframe.py
```diff
@@ -252,11 +252,95 @@ def min(self, numeric_only=True):
...
+            The value of the variance for each numeric column
+
+        See Also
+        --------
+        :pandas_api_docs:`pandas.DataFrame.var`
+
+        Examples
+        --------
+        >>> df = ed.DataFrame('localhost', 'flights')
+        >>> df.std() # doctest: +SKIP
```
sethmlarson, line 270: Should be .var()?

mesejo: Done

mesejo:
@sethmlarson Regarding the refactoring, here are my thoughts:

At least `aggs`, `_metric_aggs`, and `describe` share functionality. All of the functions built aggregation queries, perform some operations, and extract the information from the aggregation. My suggestion is to extract the built aggregation and extraction functionalities from those functions into separate functions (some candidates): `built_agg_query` and `extract_agg_response` and refactor the above mentioned functions.

sethmlarson:
@mesejo Would be good to think about `terms_aggs()` and `hist_aggs()` as well, I like the idea of build_X_query() and extract_X_response(). Would be good on the `extract_X_response()` to have the return type be something that can be returned into a `pd.DataFrame()` or `pd.Series()` so it can be used within `agg()` which returns `DataFrame` and the individual aggs like `median()` that return a `Series`.

mesejo:
@sethmlarson I'm willing to do it, but I cannot guarantee that I can deliver it fast

Issue #179, opened 2020-04-14
Refactor Aggregations to use same code path regardless of end container


sethmlarson:
Discussed this issue in this comment: https://github.com/elastic/eland/pull/175#pullrequestreview-391879466

Ideally we'd like a way to construct agg queries and unpack them and maintain that logic all in one function rather than multiple places.

sethmlarson:
Closed in #189

PR #185, opened 2020-04-15
Fixing var, std and flaky tests

### LINKED ISSUE #182: Fix .var() to use Sample variance instead of Population
Looks like one of the aggregations in the extended_metrics test suite is failing sometimes when the values are different by 11%.

stevedodson:
Closes #182

### DIFF
--- a/eland/operations.py
+++ b/eland/operations.py
@@ -287,6 +287,11 @@ class Operations:
                             results[field] = response["aggregations"][
                                 "percentiles_" + field
                             ]["values"]["50.0"]
+
+                            # If 0-length dataframe we get None here
+                            # TODO investigate if we should resolve in elasticsearch-py issue?
+                            if results[field] is None:
+                                results[field] = np.float64(np.NaN)
                         elif func[1] == "variance":
                             # pandas computes the sample variance
                             # Elasticsearch computes the population variance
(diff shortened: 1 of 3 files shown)


stevedodson:
https://github.com/elastic/eland/issues/182

eland/operations.py
```diff
@@ -285,6 +287,45 @@ def _metric_aggs(
                             results[field] = response["aggregations"][
                                 "percentiles_" + field
                             ]["values"]["50.0"]
+                        elif func[1] == "variance":
```
sethmlarson, line 290: We'll also need to add these transformations into `DataFrame.agg()` until #179 is resolved. Definitely feeling the pain until that refactor.

eland/operations.py
```diff
@@ -285,6 +287,45 @@ def _metric_aggs(
...
+                                "count"
+                            ]
+
+                            results[field] = response["aggregations"][
+                                func[0] + "_" + field
+                            ][func[1]]
+
+                            # transform population variance into sample variance
+                            if count == 0:
+                                results[field] = pd.NaN
```
sethmlarson, line 303: Can we have test cases for var+std that hits the `pd.NaN` case

stevedodson: ++ - many thanks for spotting, `pd.NaN` is incorrect + test fails

eland/operations.py
stevedodson: ++ on refactor - getting pretty messy

eland/operations.py
```diff
@@ -285,6 +287,50 @@ def _metric_aggs(
                             results[field] = response["aggregations"][
                                 "percentiles_" + field
                             ]["values"]["50.0"]
+
+                            # If 0-length dataframe we get None here
+                            # TODO investigate if we should resolve in elasticsearch-py issue?
```
sethmlarson, line 292: I'm not sure if this can be solved in elasticsearch-py? The JSON result is `null` if there are no docs in the aggregation so we'll have to handle it here.

PR #187, opened 2020-04-18
[WIP] Refactor building and extracting functionalities

mesejo:
This mockup is the result from the discussion  in #176 

#### Context
Currently the function `_metric_aggs` has many responsibilities, build the query, extract the response, transform the results to a pandas compatible format. 

#### Solution (partial)
Extract away the functionality of building and extracting from `_metric_aggs` into two functions:
- `_build_agg_query`, builds the aggregation query from a list of functions, fields
- `_extract_agg_response` extracts a dictionary (can be used in a Series or DataFrame), from the response

Additionally functions for computing the sample variance (std) were added. 

#### Notes
Perhaps it could be interesting to always use tuples as function names, this way a uniform field access can be created to all functions. Examples:

- `('percentiles', 'values', '50%')` for median
- `('extendend_stats', 'std', 'value')` for std
- `('', 'mean', 'value')` for mean 

#### Observations
What is the purpose of the `aggregatable` parameter? Seems that is only used on `nunique`

### DIFF
--- a/eland/operations.py
+++ b/eland/operations.py
@@ -167,6 +167,25 @@ class Operations:
     def hist(self, query_compiler, bins):
         return self._hist_aggs(query_compiler, bins)
 
+    def _extract_agg_response(self, response, funcs, field_names):
+        result = {}
+        for func in funcs:
+            for field_name in field_names:
+
+                if isinstance(func, tuple):
+                    func_name, argument_name = func
+                    if func_name == "percentiles":
+                        result[field_name] = response["aggregations"][
+                            f"{func_name}_{field_name}"
+                        ]["values"][argument_name]
+                    else:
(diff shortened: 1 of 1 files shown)


sethmlarson:
Thanks for this!! I have a PR that makes a lot of changes to this function too locally. I can push it up then we can look at both. We'll probably want a combination of both of these PRs in the end :)

sethmlarson:
@mesejo Here's that draft PR: https://github.com/elastic/eland/pull/189

sethmlarson:
jenkins test this please

mesejo:
@sethmlarson The PR #189 looks awesome, I agree with your idea of putting more smarts into the fields.

sethmlarson:
I'm going to close this since #179 has landed now, thanks @mesejo!

PR #189, opened 2020-04-18
Add Field from FieldMapping with all info on field

sethmlarson:
This PR still needs a little work, putting it up early so @mesejo can see what I'm up to.
Basically the thought process is we're using FIeldMapping to store a lot of information but the APIs only allow access to a subset of the information per function call, instead we should focus on putting a lot of the "smarts" in the `Field` class so that it's all localized (like which aggregations can be used on a certain Field, etc)

I'd like to continue moving the return types of `FieldMapping` to be mostly just `Field`s.

This also has the plus of making it easy to fill in `NaN` whenever an aggregation isn't possible for ES in the same way that Pandas does this (like calling `.mean()` on a string type)

Still need to add a bunch of tests for this functionality.

### DIFF
--- a/eland/series.py
+++ b/eland/series.py
@@ -1222,7 +1222,7 @@ class Series(NDFrame):
         Examples
         --------
         >>> s = ed.DataFrame('localhost', 'flights')['AvgTicketPrice']
-        >>> s.var().astype(int)
+        >>> int(s.var())
         70964
         """
         results = super().var(numeric_only=numeric_only)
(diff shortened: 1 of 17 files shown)


eland/series.py
```diff
@@ -1276,8 +1276,8 @@ def mad(self, numeric_only=None):
         Examples
         --------
         >>> s = ed.DataFrame('localhost', 'flights')['AvgTicketPrice']
-        >>> s.mad()
-        213.40706409272698
+        >>> int(s.mad())
```
stevedodson, line 1279: nice idea to make doctests more robust!

PR #191, opened 2020-04-20
Update ScriptFilter

mesejo:
The body of ScriptFilter was change to source and lang="painless" was added as explicit parameter, the test were updated to reflect the changes made both to ScriptFilter and the call of the constructor.


sethmlarson:
Jenkins test this please

PR #205, opened 2020-05-04
Fix bug with empty dataframe number of rows

### LINKED ISSUE #152: Empty DataFrames don't have proper number of rows after .head()/.tail()
This is caused by the `if size is not None` within `Operations.index_count()`, should we add a parameter to ignore `if size is not None` logic and ignore within `Index.__len__`? Would create a request per call to `__len__` though. 

```python
>>> df = ed.DataFrame(es, 'flights')
>>> print(df[df["Carrier"] == "nada"].tail().shape)
(5, 27)
>>> df = df._to_pandas()
>>> print(df[df["Carrier"] == "nada"].tail().shape)
(0, 27)
```

mesejo:
This is an attempt at solving the bug with the wrong shape, instead of only passing the sort field, one could pass the index and query the `len` in the constructor. An enhancement will be to create a unique constructor in SizeTask with the logic.

Closes #152

### DIFF
--- a/eland/tests/dataframe/test_head_tail_pytest.py
+++ b/eland/tests/dataframe/test_head_tail_pytest.py
@@ -6,6 +6,7 @@
 
 from eland.tests.common import TestData
 from eland.tests.common import assert_pandas_eland_frame_equal
+from eland.utils import eland_to_pandas
 
 
 class TestDataFrameHeadTail(TestData):
@@ -95,3 +96,9 @@ class TestDataFrameHeadTail(TestData):
         df = df[["timestamp", "OriginAirportID", "DestAirportID", "FlightDelayMin"]]
         df = df.tail()
         print(df)
+
+    def test_doc_test_tail_empty(self):
+        df = self.ed_flights()
+        df = df[df.OriginAirportID == "NADA"]
+        df = df.tail()
+        assert df.shape[0] == 0
--- a/eland/tests/dataframe/test_head_tail_pytest.py
+++ b/eland/tests/dataframe/test_head_tail_pytest.py
@@ -6,7 +6,6 @@
 
 from eland.tests.common import TestData
 from eland.tests.common import assert_pandas_eland_frame_equal
-from eland.utils import eland_to_pandas
 
 
 class TestDataFrameHeadTail(TestData):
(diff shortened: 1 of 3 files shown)


sethmlarson:
jenkins test this please

sethmlarson:
@mesejo Nice! Let me know when you're ready for review. :)

mesejo:
@sethmlarson For my part the only thing I think is missing is refactoring the SizeTask constructor, given that all the class children share the same behavior. Let me know what do you think.

sethmlarson:
@mesejo That sounds fine to me!

mesejo:
@sethmlarson Feel free to review it when possible

eland/tests/dataframe/test_head_tail_pytest.py
```diff
@@ -6,6 +6,7 @@
 
 from eland.tests.common import TestData
 from eland.tests.common import assert_pandas_eland_frame_equal
+from eland.utils import eland_to_pandas
```
sethmlarson, line 9: Do we need this import?

sethmlarson:
jenkins test this please

eland/tests/dataframe/test_head_tail_pytest.py
```diff
@@ -95,3 +96,9 @@ def test_doc_test_tail(self):
         df = df[["timestamp", "OriginAirportID", "DestAirportID", "FlightDelayMin"]]
         df = df.tail()
         print(df)
+
+    def test_doc_test_tail_empty(self):
+        df = self.ed_flights()
+        df = df[df.OriginAirportID == "NADA"]
+        df = df.tail()
+        assert df.shape[0] == 0
```
sethmlarson, line 104: Can we add a case for n=1 as well?

mesejo: By n=1 you mean `df.tail(1)` ?

sethmlarson: Oh sorry I meant a case where the dataframe contains 1 row so calling `df.tail().shape[0] == 1`

mesejo: Done!

PR #220, opened 2020-06-15
Error when ingest.simulate fails without non-2XX HTTP status code

sethmlarson:
- Adds support for submitting uncompressed ML model definitions. This helps if users want to receive errors immediately when creating a model over the compressed definition which errors out on first usage. Eland also happened to find an Elasticsearch bug by using the compressed definition so it'd be good to test both.
- Changes `ImportedMLModel.predict()` to error out if an `error` is received within `docs` without a non-2XX HTTP status code.


sethmlarson:
CI is failing for an unrelated reason, going to merge this.

Issue #222, opened 2020-07-01
(Feature) Dataframe column autocompletion


leonardbinet:
Nice library 👍 

One nice pandas feature is to provide autocompletion on column names in interactive environment:

<img width="646" alt="Capture d’écran 2020-07-01 à 08 52 26" src="https://user-images.githubusercontent.com/1590371/86212412-45382b00-bb78-11ea-9cdd-a531b5be6ba5.png">

It could be useful to provide similar experience in eland.

PR #223, opened 2020-07-01
Dataframe column names autocompletion

leonardbinet:
Closes https://github.com/elastic/eland/issues/222

### DIFF
--- a/eland/dataframe.py
+++ b/eland/dataframe.py
@@ -25,7 +25,7 @@ from eland.ndframe import NDFrame
 from eland.series import Series
 from eland.common import DEFAULT_NUM_ROWS_DISPLAYED, docstring_parameter
 from eland.filter import BooleanFilter
-from eland.utils import deprecated_api
+from eland.utils import deprecated_api, is_valid_attr_name
 
 
 class DataFrame(NDFrame):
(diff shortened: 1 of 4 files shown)


eland/dataframe.py
```diff
@@ -445,6 +445,16 @@ def drop(
     def __getitem__(self, key):
         return self._getitem(key)
 
+    def __dir__(self):
+        """
+        Provide autocompletion on field names in interactive environment.
+        """
+        return super(NDFrame, self).__dir__() + [
```
sethmlarson, line 452: Is this intentionally `super(NDFrame, self)` instead of `super().__dir__()`?

sethmlarson:
jenkins test this please

eland/utils.py
```diff
@@ -23,3 +24,10 @@ def wrapped(*args, **kwargs):
         return wrapped
 
     return wrapper
+
+
+def is_valid_attr_name(s):
+    """
+    Ensure the given string can be used as attribute on an object instance.
+    """
+    return isinstance(s, str) and re.search(string=s, pattern=r"^[a-zA-Z_]+[a-zA-Z0-9_]*$") and s.replace('_', '')
```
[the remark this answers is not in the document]
leonardbinet: - you're right on regex, the `+` is actually useless
- I didn't think `__` was a valid attribute name but it is indeed :) so I removed the last condition

eland/dataframe.py
leonardbinet: I was writing in p2/3 compatible, but indeed unnecessary here since eland is designed for python3

sethmlarson:
jenkins test this please

PR #230, opened 2020-07-13
Use SymmetricalAPIChecker for testing same behavior/results

sethmlarson:
The object maintains an Eland object and Pandas object which are initially a DataFrame. The object then responds to all API calls by applying the call to each object and comparing the results. If the results are both Pandas objects or literals then we stop tracking otherwise the tracking continues. This lets us write shorter test cases and exercise similarities between the two APIs more easily.

If this is desirable it can no doubt be applied to more test cases.


sethmlarson:
Failing on `script.max_compilation_rate` being remove in 8.0

Issue #242, opened 2020-07-17
Add multiclass classification support in model transformers


benwtrent:
With the release of multi-class classification in Elasticsearch v7.7.0, we now support model formats with more than one leaf value. 

Investigations and changes to the supported models should be made to more fully support multiclass classification.

Example multi-class tree:
```
"tree_structure" : [
                    {
                      "decision_type" : "lt",
                      "threshold" : 4.806131127865606,
                      "split_feature" : 0,
                      "split_gain" : 13.17420833661845,
                      "node_index" : 0,
                      "default_left" : true,
                      "left_child" : 1,
                      "right_child" : 2,
                      "number_samples" : 150
                    },
                    {
                      "decision_type" : "lt",
                      "threshold" : 0.5266284997189365,
                      "split_feature" : 1,
                      "split_gain" : 10.014818498134217,
                      "node_index" : 1,
                      "default_left" : true,
                      "left_child" : 3,
                      "right_child" : 4,
                      "number_samples" : 99
                    },
                    {
                      "decision_type" : "lte",
                      "node_index" : 2,
                      "leaf_value" : [
                        -0.03783979371620798,
                        -0.006144227496540516,
                        0.043984021212748546
                      ],
                      "default_left" : false,
                      "number_samples" : 51
                    },
                    {
                      "decision_type" : "lte",
                      "node_index" : 3,
                      "leaf_value" : [
                        0.05803486743667319,
                        -0.029217359098602222,
                        -0.02881750833807117
                      ],
                      "default_left" : false,
                      "number_samples" : 49
                    },
                    {
                      "decision_type" : "lte",
                      "node_index" : 4,
                      "leaf_value" : [
                        -0.038383229272749654,
                        0.04458938677045973,
                        -0.006206157497710134
                      ],
                      "default_left" : false,
                      "number_samples" : 50
                    }
                  ],
```

Elastic ML's internal output aggregator is a multi-class logistic regression. 

The calculated leaves are summed up on axis 1 and then a `softmax` calculation is made. 

This type of calculation might not work out of the box for certain multi-class tree models.

PR #246, opened 2020-08-05
[ML] add multi:softmax|softprob XGBoost support

benwtrent:
This commit adds support for XGBoost multi-class classification model transformations.

XGBoost handles multiclass classification by having `n_classes * n_estimators` trees. 
Then, every `n_classes` tree corresponds to their respective classes

Since Elasticsearch supports multi-valued leaves, we can transform the xgboost format by choosing the
appropriate leaf value index given the tree id.

This commit also fixes a minor model transformation bug where if xgboost actually defined the feature_names instead of the default `f1,...fn` we would blow up. 

closes https://github.com/elastic/eland/issues/242

### DIFF
--- a/eland/tests/ml/test_imported_ml_model_pytest.py
+++ b/eland/tests/ml/test_imported_ml_model_pytest.py
@@ -228,8 +228,10 @@ class TestImportedMLModel:
     @pytest.mark.parametrize("compress_model_definition", [True, False])
     def test_xgb_classifier(self, compress_model_definition):
         # Train model
-        training_data = datasets.make_classification(n_features=5)
-        classifier = XGBClassifier(booster="gbtree")
+        training_data = datasets.make_classification(
+            n_features=5, n_classes=3, n_informative=3
+        )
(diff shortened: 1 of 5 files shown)


eland/tests/ml/test_imported_ml_model_pytest.py
```diff
@@ -228,8 +228,10 @@ def test_random_forest_regressor(self, compress_model_definition):
     @pytest.mark.parametrize("compress_model_definition", [True, False])
     def test_xgb_classifier(self, compress_model_definition):
         # Train model
-        training_data = datasets.make_classification(n_features=5)
-        classifier = XGBClassifier(booster="gbtree")
+        training_data = datasets.make_classification(
```
sethmlarson, line 231: Can we add an additional test (or parametrize this test) so we maintain both the binary and multi classification?

benwtrent: @sethmlarson added a new parameter :)

Issue #240, opened 2020-07-17
Add support for LightGBM machine learning models


sethmlarson:
https://lightgbm.readthedocs.io/en/latest/Python-API.html

sethmlarson:
Closed in #247 and #252

PR #247, opened 2020-08-06
[ML] adds support for transforming LGBMRegressor models

benwtrent:
This adds support for lightgbm LGBMRegressor models.

Verified that all the supported objective functions and various booster types work great.

relates to https://github.com/elastic/eland/issues/240


eland/ml/imported_ml_model.py
```diff
@@ -59,14 +63,28 @@ class ImportedMLModel(MLModel):
...
+                - "huber"
+                - "fair"
+                - "quantile"
+                - "mape"
+            - The following booster types are supported
+                - "gbdt"
+                - "rf"
+                - "dart"
+                - "goss"
+            - Categorical fields are expected to already be processed
```
sethmlarson, line 79: I wonder if it makes sense to have this caveat at the top of the section so it isn't visually mistaken for a list element?

eland/ml/transformers/__init__.py
```diff
@@ -82,3 +82,16 @@ def get_model_transformer(model: Any, **kwargs: Any) -> ModelTransformer:
     _MODEL_TRANSFORMERS.update(_XGBOOST_MODEL_TRANSFORMERS)
 except ImportError:
     pass
+
+try:
+    from .lightgbm import (
+        LGBMRegressor,
+        LGBMForestTransformer,
+        LGBMRegressorTransformer,
+        _MODEL_TRANSFORMERS as _LGBOOST_MODEL_TRANSFORMERS,
```
sethmlarson, line 91: nit: `... as _LIGHTGBM_MODEL_TRANSFORMERS`

eland/ml/transformers/lightgbm.py
```diff
@@ -0,0 +1,202 @@
...
+
+class Counter:
+    def __init__(self, start: int = 0):
+        self._value = start
+
+    def inc(self) -> int:
+        self._value += 1
+        return self._value
+
+    def value(self) -> int:
```
sethmlarson, line 51: I don't think the `value()` method is used, potentially remove?

eland/ml/transformers/lightgbm.py
```diff
@@ -0,0 +1,202 @@
...
+        self._node_decision_type = "lte"
+        self._objective = model.params["objective"]
+
+    def build_tree(self, tree_json_obj: Dict[str, Any]) -> Tree:
+        tree_nodes = list()
+        next_id = Counter()
+        py_id_to_node_id: Dict[int, int] = dict()
+
+        def add_tree_node(tree_node_json_obj: Dict[str, Any], counter: Counter) -> None:
+            curr_id = py_id_to_node_id[id(tree_node_json_obj)]
```
sethmlarson, line 81: I'll just note that the use of `id()` odd, is there no other way to identify tree nodes? That's unfortunate :(

eland/ml/transformers/lightgbm.py
```diff
@@ -0,0 +1,202 @@
...
+            if parse_left:
+                add_tree_node(tree_node_json_obj["left_child"], counter)
+            if parse_right:
+                add_tree_node(tree_node_json_obj["right_child"], counter)
+
+        py_id_to_node_id[id(tree_json_obj["tree_structure"])] = next_id.value()
+        add_tree_node(tree_json_obj["tree_structure"], next_id)
+        tree_nodes.sort(key=lambda n: n.node_idx)
+        return Tree(
+            self._feature_names,
```
sethmlarson, line 123: nit: use the `feature_names=` kwarg too

eland/tests/ml/test_imported_ml_model_pytest.py
```diff
@@ -50,6 +57,10 @@
     reason="This test requires 'scikit-learn' and 'xgboost' to not be installed",
 )
 
+requires_lgboost = pytest.mark.skipif(
```
sethmlarson, line 60: nit: change the name to `..._lightgbm` here too

eland/ml/transformers/lightgbm.py
benwtrent: There probably is, but I did not want to bother with it any longer. 

The underlying numbering scheme that lightgbm uses does not follow our convention. 

I could maybe fiddle with it more to effectively build the node list backwards with recursion, but this way seemed simple enough.

eland/ml/transformers/lightgbm.py
benwtrent: It is used now :D

eland/ml/transformers/lightgbm.py
benwtrent: I went ahead and used the `Counter` to mutate the node ids and recursively added nodes backwards (depth first). This was simple enough because I was already sorting the node list at the end :)

eland/ml/transformers/lightgbm.py
```diff
@@ -0,0 +1,188 @@
...
+    if decider == "<=":
+        return "lte"
+    if decider == "<":
+        return "lt"
+    if decider == ">":
+        return "gt"
+    if decider == ">=":
+        return "gte"
+    raise ValueError(
+        "Unsupported splitting decider: %s. Only <=, " "<, >=, and > are allowed."
```
sethmlarson, line 39: This is two string literals joining together, you can change it to one. (see `"...<=, " "<,..."`)

eland/ml/imported_ml_model.py
```diff
@@ -59,14 +63,28 @@ class ImportedMLModel(MLModel):
...
+        - lightgbm.LGBMRegressor
+            - Categorical fields are expected to already be processed
+            - Only the following objectives are supported
+                - "regression"
+                - "regression_l1"
+                - "huber"
+                - "fair"
+                - "quantile"
+                - "mape"
+            - The following booster types are supported
```
tveasey, line 75: I'm not sure I'd bother to mention booster types here. This is a training detail and so doesn't affect inference.

eland/ml/transformers/lightgbm.py
```diff
@@ -0,0 +1,188 @@
...
+    warning: do not use directly. Use a derived classes instead
+    """
+
+    def __init__(
+        self,
+        model: Booster,
+        feature_names: List[str],
+        classification_labels: Optional[List[str]] = None,
+        classification_weights: Optional[List[float]] = None,
+    ):
```
tveasey, line 68: Why not?
```suggestion
        classification_weights: Optional[List[float]] = None):
```

eland/ml/transformers/lightgbm.py
```diff
@@ -0,0 +1,188 @@
...
+        tree_nodes = list()
+        next_id = Counter()
+
+        def add_tree_node(tree_node_json_obj: Dict[str, Any], counter: Counter) -> int:
+            curr_id = counter.value()
+            if "leaf_value" in tree_node_json_obj:
+                tree_nodes.append(
+                    TreeNode(
+                        node_idx=curr_id,
+                        leaf_value=[float(tree_node_json_obj["leaf_value"])],
```
tveasey, line 85: nit: I know this is harmless but I find including the trailing comma distracting:
```suggestion
                        leaf_value=[float(tree_node_json_obj["leaf_value"])]
```
There are a few cases like this.

eland/tests/ml/test_imported_ml_model_pytest.py
```diff
@@ -322,3 +333,35 @@ def test_predict_single_feature_vector(self):
 
         # Clean up
         es_model.delete_model()
+
+    @requires_lightgbm
+    @pytest.mark.parametrize("compress_model_definition", [True, False])
+    def test_lgbm_regressor(self, compress_model_definition):
```
tveasey, line 339: I think this ought really to be testing `LGBMClassifier` as well.

eland/ml/imported_ml_model.py
```diff
@@ -59,14 +63,28 @@ class ImportedMLModel(MLModel):
...
         - xgboost.XGBClassifier
-            - only the following operators are supported:
+            - only the following objectives are supported:
                 - "binary:logistic"
                 - "binary:hinge"
                 - "multi:softmax"
                 - "multi:softprob"
         - xgboost.XGBRegressor
-            - only the following operators are supportd:
+            - only the following objectives are supported:
```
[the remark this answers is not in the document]
benwtrent: @tveasey I think the objective can change what values are stored in the leaves. Which might change how inference is done... (storing log likelihood, etc.)

eland/ml/imported_ml_model.py
benwtrent: I suppose, but we do check the booster type to make sure it is a model type that we actually support. Example, XGBoost supports a linear regression booster that results in a completely different format.

eland/ml/transformers/lightgbm.py
benwtrent: autoformatting

eland/tests/ml/test_imported_ml_model_pytest.py
benwtrent: @tveasey support for that class will happen later. This PR is regression only (to keep size small)

eland/ml/imported_ml_model.py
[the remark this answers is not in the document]
tveasey: It can, but does it for say `pseudohubererror`. I'd be surprised if it did, but maybe it does for some unexpected reason. Did you test this list in the past?

eland/ml/imported_ml_model.py
tveasey: Ok gotcha. I see now see the source of this. This is unfortunate because different concepts are conflated as booster between the two libraries. What xgboost labels booster is really the base learner as opposed to what GBM calls booster, which does actually select the boosting algorithm. In that case, I'd be inclined to somehow disambiguate the concepts in the converters.

eland/ml/transformers/lightgbm.py
sethmlarson: also re this behavior of trailing commas, it makes adding additional parameters to a function only effect one line in the Git history. Without the comma the diff would touch two lines: one to add the parameter and one adding the comma on the line before.

tveasey: Maybe this is more idiomatic python, it just looks weird to me, i.e. I find my self wondering what was missed.

eland/tests/ml/test_imported_ml_model_pytest.py
tveasey: And indeed you haven't implemented the sigmoid transforms. Good point!

eland/ml/imported_ml_model.py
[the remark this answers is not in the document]
benwtrent: @tveasey I did not test it. 

In a separate PR I can add some tests that cycle through objectives to verify that their support (and catch regressions).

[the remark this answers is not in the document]
tveasey: Ok great. This is really just correcting a typo so agreed no need to worry about it in this PR, just thought I'd ask the question while it occurred to me.

sethmlarson:
Thanks @benwtrent and @tveasey!

PR #249, opened 2020-08-10
Replace MLModel(overwrite) with es_if_exists

### LINKED ISSUE #225: Replace MLModel(overwrite) with es_if_exists
Mirrors the `pandas_to_eland()` API

V1NAY8:
Closes #225 
- Added es_if_exists in ImportedMLModel and deprecated the overwrite parameter.
- Updated the tests of ImportedMLModel.
- Updated Contributing guidelines for installation instructions using Docker for helping new contributors.
- Updated info_es() to es_info() (because info_es() is deprecated)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -151,7 +151,8 @@ currently using a minimum version of PyCharm 2019.2.4.
     Tools\'-\>\'Docstring format\' to `numpy`
 -   Install development requirements. Open terminal in virtual
     environment and run `pip install -r requirements-dev.txt`
--   Setup Elasticsearch instance (assumes `localhost:9200`), and run
+-   Setup Elasticsearch instance with docker `ELASTICSEARCH_VERSION=elasticsearch:7.x-SNAPSHOT .ci/run-elasticsearch.sh` and check `http://localhost:9200`
+- Run
     `python -m eland.tests.setup_tests` to setup test environment -*note
     this modifies Elasticsearch indices*
 -   Install local `eland` module (required to execute notebook tests)
--- a/eland/ml/imported_ml_model.py
+++ b/eland/ml/imported_ml_model.py
@@ -101,10 +101,13 @@ class ImportedMLModel(MLModel):
     classification_weights: List[str]
         Weights of the classification targets
 
-    es_if_exists: bool
-        Delete and overwrite existing model (if exists)
-    
-    overwrite:**DEPRECATED** - bool
+    es_if_exists: {'fail', 'replace} default 'fail'
+        How to behave if model already exists
+
+        - fail: Raise a Value Error
+        - replace: Overwrite existing model
+
+    overwrite: **DEPRECATED** - bool
         Delete and overwrite existing model (if exists)
 
     es_compress_model_definition: bool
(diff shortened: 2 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -151,7 +151,8 @@ currently using a minimum version of PyCharm 2019.2.4.
     Tools\'-\>\'Docstring format\' to `numpy`
 -   Install development requirements. Open terminal in virtual
     environment and run `pip install -r requirements-dev.txt`
--   Setup Elasticsearch instance (assumes `localhost:9200`), and run
+-   Setup Elasticsearch instance with docker `ELASTICSEARCH_VERSION=elasticsearch:7.x-SNAPSHOT .ci/run-elasticsearch.sh` and check `http://localhost:9200`
```
sethmlarson, line 154: Thanks for this!

eland/ml/imported_ml_model.py
```diff
@@ -150,7 +155,15 @@ def __init__(
         self._model_type = transformer.model_type
         serializer = transformer.transform()
 
+        if es_if_exists:
+            self.delete_model()
+
         if overwrite:
```
sethmlarson, line 161: We should error out with `ValueError` if both `overwrite` and `es_if_exists` are given.

eland/ml/imported_ml_model.py
```diff
@@ -150,7 +155,15 @@ def __init__(
         self._model_type = transformer.model_type
         serializer = transformer.transform()
 
+        if es_if_exists:
```
sethmlarson, line 158: We should be checking `if es_if_exists == "replace"`, not for truthiness

eland/ml/imported_ml_model.py
V1NAY8: Please help me out. I have completed the logic but facing one issue
**How can we check if both are passed during the object creation?**
Overwrite can take None/True/False
I tried many combinations: Either raising valueerror is failing or the below logic isn't getting executed.
```python
# Since we are  defaulting overwrite  = None and es_if_exists = 'fail'
    if not (isinstance(overwrite,type(None)) and es_if_exists in ['fail','replace']):
        raise ValueError("Use es_if_exists(params 'fail' or 'replace') or overwrite(**Deprecated**).")
# Logic
if not overwrite: 
    es_if_exists = 'replace'

if es_if_exists == 'fail':
   # raise error by checking.
elif es_if_exists == 'replace'
   # delete model
```

V1NAY8: Can we replace the overwrite with es_if_exists to avoid complexity?

sethmlarson: This is what I'd do:

```python
def __init__(self, overwrite=None, es_if_exists=None):
    if overwrite is not None and es_if_exists is not None:
        raise ValueError("Using 'overwrite' and 'es_if_exists' together is invalid, use only 'es_if_exists'")
    if overwrite is not None:
        warnings.warn(
            f"'overwrite' parameter is deprecated, use 'es_if_exists' instead",
            DeprecationWarning,
        )
        es_if_exists = "replace" if overwrite else "fail"
    elif es_if_exists is None:
        es_if_exists = "fail"
    if es_if_exists not in ("fail", "replace"):
        raise ValueError("'es_if_exists' must be either 'fail' or 'replace'")
    
```

sethmlarson:
Jenkins test this please

V1NAY8:
Jenkins test this please

sethmlarson:
Jenkins test this please

sethmlarson:
It looks like Mypy is failing, could you try running `nox -s lint` locally:

From Jenkins:

```
07:23:16 eland/ml/imported_ml_model.py:165: error: Incompatible default for argument "es_if_exists" (default has type "None", argument has type "bool")
07:23:16 eland/ml/imported_ml_model.py:198: error: Non-overlapping container check (element type: "bool", container item type: "str")
07:23:16 eland/ml/imported_ml_model.py:200: error: Non-overlapping equality check (left operand type: "bool", right operand type: "Literal['fail']")
07:23:16 eland/ml/imported_ml_model.py:205: error: Non-overlapping equality check (left operand type: "bool", right operand type: "Literal['replace']").
```

sethmlarson:
jenkins test this please

V1NAY8:
Thanks @sethmlarson for the inputs and help. Will Continue to contribute 😃

PR #250, opened 2020-08-11
[ML] testing supported objectives and boosters

benwtrent:
Adds testing coverage and fixes some latent bugs in model transformers.

### DIFF
--- a/eland/ml/imported_ml_model.py
+++ b/eland/ml/imported_ml_model.py
@@ -75,7 +75,6 @@ class ImportedMLModel(MLModel):
         - xgboost.XGBClassifier
             - only the following objectives are supported:
                 - "binary:logistic"
-                - "binary:hinge"
                 - "multi:softmax"
                 - "multi:softprob"
         - xgboost.XGBRegressor
@@ -84,6 +83,7 @@ class ImportedMLModel(MLModel):
                 - "reg:linear"
                 - "reg:squaredlogerror"
                 - "reg:logistic"
+                - "reg:pseudohubererror"
 
     feature_names: List[str]
         Names of the features (required)
--- a/eland/tests/ml/test_imported_ml_model_pytest.py
+++ b/eland/tests/ml/test_imported_ml_model_pytest.py
@@ -365,10 +365,19 @@ class TestImportedMLModel:
         "objective",
         ["regression", "regression_l1", "huber", "fair", "quantile", "mape"],
     )
-    def test_lgbm_regressor(self, compress_model_definition, objective):
+    @pytest.mark.parametrize("booster", ["gbdt", "rf", "dart", "goss"])
+    def test_lgbm_regressor(self, compress_model_definition, objective, booster):
         # Train model
         training_data = datasets.make_regression(n_features=5)
-        regressor = LGBMRegressor(objective=objective)
+        if booster == "rf":
+            regressor = LGBMRegressor(
+                boosting_type=booster,
+                objective=objective,
+                bagging_fraction=0.5,
+                bagging_freq=3,
+            )
+        else:
+            regressor = LGBMRegressor(boosting_type=booster, objective=objective)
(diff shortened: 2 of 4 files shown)


eland/ml/imported_ml_model.py
```diff
@@ -75,7 +75,6 @@ class ImportedMLModel(MLModel):
         - xgboost.XGBClassifier
             - only the following objectives are supported:
                 - "binary:logistic"
-                - "binary:hinge"
```
sethmlarson, line 78: Is there anything we can do to support "binary:hinge"? If so we could add an issue to implement support.

eland/tests/ml/test_imported_ml_model_pytest.py
```diff
@@ -62,6 +62,13 @@
 )
 
 
+def check_prediction_equality(es_model, py_model, test_data):
```
sethmlarson, line 65: This is a great change! :tada:

eland/tests/ml/test_imported_ml_model_pytest.py
```diff
@@ -268,24 +256,62 @@ def test_xgb_classifier(self, compress_model_definition, multi_class):
...
+        check_prediction_equality(es_model, classifier, test_data)
+
+        # Clean up
+        es_model.delete_model()
+
+    @requires_xgboost
+    @pytest.mark.parametrize("compress_model_definition", [True, False])
+    @pytest.mark.parametrize(
+        "objective",
+        ["reg:squarederror", "reg:squaredlogerror", "reg:linear", "reg:logistic"],
```
sethmlarson, line 303: Should we add "reg:pseudohubererror" here too?

eland/tests/ml/test_imported_ml_model_pytest.py
```diff
@@ -358,10 +383,38 @@ def test_lgbm_regressor(self, compress_model_definition):
...
+        test_data = [[0.1, 0.2, 0.3, -0.5, 1.0], [1.6, 2.1, -10, 50, -1.0]]
+        check_prediction_equality(es_model, regressor, test_data)
 
-        es_results = es_model.predict(test_data)
+        # Clean up
+        es_model.delete_model()
 
-        np.testing.assert_almost_equal(test_results, es_results, decimal=2)
+    @requires_lightgbm
+    @pytest.mark.parametrize("booster", ["gbdt", "rf", "dart", "goss"])
```
sethmlarson, line 394: I don't know a lot here but is it possible to move this `parametrize()` call to the above test suite or if that'd be too much effort to test all permutations? I don't know much about LightGBM so I'm unsure if this request even makes sense :)

eland/ml/imported_ml_model.py
benwtrent: I can try some more, but all the different variants failed periodically. One would think since leaf values are `0/1` that our typical prediction logic would work :(

eland/tests/ml/test_imported_ml_model_pytest.py
benwtrent: @sethmlarson while we do support it, our current version of XGBoost does not include it :D

eland/ml/imported_ml_model.py
benwtrent: hinge won't work without some backend changes :(

eland/tests/ml/test_imported_ml_model_pytest.py
benwtrent: will move it

eland/tests/ml/test_imported_ml_model_pytest.py
benwtrent: XGBoost will include this in 1.2. I think we can add an automated test at that point.

sethmlarson:
Opened https://github.com/elastic/eland/issues/251 to track the addition of testing for `reg:pseudohubererror`

PR #252, opened 2020-08-11
[ML] add LGBMClassifier transform support

benwtrent:
Adds support for transforming LGBMClassifier models. 

We are able to support binary, multiclass, and multiclassova. These are the only valid ones for classification. 
But, still restricting to prevent a regression where lightgbm adds a new objective without us knowing.

### DIFF
--- a/eland/tests/ml/test_imported_ml_model_pytest.py
+++ b/eland/tests/ml/test_imported_ml_model_pytest.py
@@ -39,7 +39,7 @@ except ImportError:
     HAS_XGBOOST = False
 
 try:
-    from lightgbm import LGBMRegressor
+    from lightgbm import LGBMRegressor, LGBMClassifier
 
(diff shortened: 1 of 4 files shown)


eland/tests/ml/test_imported_ml_model_pytest.py
```diff
@@ -62,6 +62,10 @@
 )
 
 
+def random_rows(data, size):
```
benwtrent, line 65: this provides MUCH better testing. 

The training data is random, and selecting a random subset to infer against provides much better coverage than a static two rows.

benwtrent:
@sethmlarson i don't think so. Much of the logic formatting is exactly the same as xgboost. The node extraction is just slightly different.

sethmlarson:
Great! I'll merge this then :)

PR #253, opened 2020-08-12
Add support for Pandas 1.1, multiple improvements for Datetime aggs

### LINKED ISSUE #244: Add support for Pandas 1.1
Looks like Pandas v1.1 changes some APIs so we'll need to restore compatibility.

sethmlarson:
- Adds support for Pandas v1.1 by moving the `try_sort` function previously deprecated in `pandas.core.common` to our own `eland.utils`
- Better precision from Datetime aggregations by using `value` instead of `value_as_string`.
  For aggregations like `mean` this was throwing away some extra precision we were getting from Elasticsearch.
- Datetime aggregations return `NaT` if the aggregation isn't applicable instead of `NaN` (this is what Pandas does)
- Fixed a bug when calling `.agg([...])` sometimes aggregations would report as unsupported for the field due to not unpacking the extended_stats tuple (see `Field.es_agg_compatible)
- Added support for `median` agg for datetimes
- Fixed a bug where `Series.median()` wasn't defined, would use essentially `DataFrame.median()` which isn't right for `Series`.
- Fixed support for Elasticsearch 7.6 where `TreeNode.leaf_values` has to be encoded as a scalar not a list
- Few misc ML tweaks that were getting masked due to our CI failing for so long

Closes #244


sethmlarson:
The 8.0.0-SNAPSHOT is failing currently due to https://github.com/elastic/elasticsearch/pull/60707, will begin passing once this lands.

eland/ml/_model_serializer.py
```diff
@@ -94,7 +94,12 @@ def to_dict(self) -> Dict[str, Any]:
             add_if_exists(d, "split_feature", self._split_feature)
             add_if_exists(d, "threshold", self._threshold)
         else:
-            add_if_exists(d, "leaf_value", self._leaf_value)
+            if len(self._leaf_value) == 1:
```
benwtrent, line 97:
```
In [2]: len(foo)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-63c8819f1adc> in <module>()
----> 1 len(foo)

TypeError: object of type 'NoneType' has no len()
```
We should protect when `self._leaf_value is None`

noxfile.py
```diff
@@ -98,25 +98,10 @@ def test(session):
     session.run("python", "-m", "eland.tests.setup_tests")
     session.run("pytest", "--doctest-modules", *(session.posargs or ("eland/",)))
 
-    session.run("python", "-m", "pip", "uninstall", "--yes", "scikit-learn", "xgboost")
+    session.run("python", "-m", "pip", "uninstall", "--yes", "lightgbm")
```
benwtrent, line 101: What makes `lightgbm` a special case here?

sethmlarson: Whoops, this should be uninstalling all the libraries to ensure they're optional. I'll fix that.

eland/ml/_model_serializer.py
sethmlarson: I believe we're guarding against `None` a few lines above so should be good here?

benwtrent: Yep, no worries here :D

PR #257, opened 2020-08-14
Update xgboost.py get_feature_id

### LINKED ISSUE #256: XGBoost issue with quick fix
Trying to upload an XGBoost model to elastic I find de following bug:

In xgboost.py in 

_def get_feature_id(self, feature_id: str) -> int:
        if feature_id[0] == "f":
            try:
                return int(feature_id[1:])
            except ValueError:
                raise RuntimeError(f"Unable to interpret '{feature_id}'")
        f_id = self._feature_dict.get(feature_id)
        if f_id **or f_id == 0**:
            return f_id
        else:
            try:
                return int(feature_id)
            except ValueError:
                raise RuntimeError(f"Unable to interpret '{feature_id}'")_

if the f_id is 0 an error raises when the name of the feature is not a number. The little code in bold can be helpful to skip the error.

jimazmarin:
Correction of a code error that raised a RunTimeError when the feature name was a String and the f_id was 0.

Closes #256

### DIFF
--- a/eland/ml/transformers/xgboost.py
+++ b/eland/ml/transformers/xgboost.py
@@ -58,7 +58,7 @@ class XGBoostForestTransformer(ModelTransformer):
             except ValueError:
                 raise RuntimeError(f"Unable to interpret '{feature_id}'")
         f_id = self._feature_dict.get(feature_id)
-        if f_id:
+        if f_id or f_id == 0:
             return f_id
         else:
             try:


sethmlarson:
This looks good to me, cc-ing @benwtrent as well.
@jimazmarin Could you add a tiny test case if possible to make sure we don't regress on this behavior?

sethmlarson:
jenkins test this please

eland/ml/transformers/xgboost.py
```diff
@@ -58,7 +58,7 @@ def get_feature_id(self, feature_id: str) -> int:
             except ValueError:
                 raise RuntimeError(f"Unable to interpret '{feature_id}'")
         f_id = self._feature_dict.get(feature_id)
-        if f_id:
+        if f_id or f_id == 0:
```
benwtrent, line 61: would `f_id is not None` work as well? I am not sure which is more pythonic.

sethmlarson: `fd_id is not None` works too :) Wasn't sure what type of falsy the value could be.

sethmlarson:
Thank you for finding this issue. I'm closing this PR in favor of https://github.com/elastic/eland/pull/261 to get a 7.9.0a1 release out sooner.

PR #263, opened 2020-08-19
DataFrame.agg() with single non-list aggregation name now returns series

### LINKED ISSUE #235: DataFrame.agg() with single non-list aggregation name returns DataFrame instead of Series
The following code produces incorrect output:

```python
from eland import DataFrame
from eland.tests.common import ES_TEST_CLIENT

df = DataFrame(ES_TEST_CLIENT, "flights")[["FlightDelayMin"]]
print("Eland")
print(df.agg("mean"))
print("Pandas")
print(df.to_pandas().agg("mean"))
```

```
Eland
      FlightDelayMin
mean       47.335171
Pandas
FlightDelayMin    47.335171
dtype: float64
```

V1NAY8:
Closes #235 
- Now non-list aggregation will return series 
- List aggregation works the same.
- Added Tests


V1NAY8:
@sethmlarson Can you please review this?

eland/dataframe.py
```diff
@@ -1386,9 +1386,12 @@ def aggregate(self, func, axis=0, *args, **kwargs):
         # ['count', 'mad', 'max', 'mean', 'median', 'min', 'mode', 'quantile',
         # 'rank', 'sem', 'skew', 'sum', 'std', 'var', 'nunique']
         if isinstance(func, str):
-            # wrap in list
-            func = [func]
-            return self._query_compiler.aggs(func)
+            if hasattr(self, func):
```
sethmlarson, line 1389: I think we can do this instead to guarantee that we only run actual aggs like `mean`, etc:

```python
        if isinstance(func, str):
            return self._query_compiler.aggs([func]).squeeze().rename(None)
```

V1NAY8: @sethmlarson That was a brilliant thought of converting dataframe to series. I was just trying to work whole around and making it complex.

I added one more edge case where ed_df.agg(["mean"]) also returns dataframe to series.

sethmlarson: Actually our previous behavior here was correct:

```python
>>> print(df.agg("mean"))
FlightTimeMin     511.127842
FlightTimeHour      8.518797
dtype: float64

>>> print(df.agg(["mean"]))
      FlightTimeMin  FlightTimeHour
mean     511.127842        8.518797
```

If there isn't one already there should be a test case for this behavior.

sethmlarson:
jenkins test this please

PR #266, opened 2020-08-21
Deprecate ImportedMLModel in favor of MLModel.import_model()

sethmlarson:
- Deprecates the separate `ImportedMLModel` class in favor of `MLModel.import_model()`.
- Moves `ImportedMLModel.predict()` to `MLModel` and allows calling predictions on already existing ML models within Elasticsearch
- Allows for using numpy arrays with `MLModel.predict()`

I wonder about the name `import_model()`, should it be something else or does this match?
Also was wondering if we one day decide to add model exporting (to a format that allows moving to other clusters) if we'd regret the name `import_` not being a perfect match with `export_`?


eland/ml/transformers/__init__.py
```diff
@@ -35,9 +35,10 @@ def get_model_transformer(model: Any, **kwargs: Any) -> ModelTransformer:
             kwargs = {k: v for k, v in kwargs.items() if k in accepted_kwargs}
 
             return transformer(model, **kwargs)
+    print(_MODEL_TRANSFORMERS)
```
benwtrent, line 38: debug?

sethmlarson: Yep! Good catch :)

stevedodson (approved):
LGTM and change makes sense. `ImportedMLModel` was legacy class added before we could do inference on Elastic ML models.

PR #271, opened 2020-09-03
Fix build errors on 8.0 snapshot

V1NAY8:
- There was a deprecation warning which says **weekyear** should be in snake case in future versions.
- Tested this on 7.x-SNAPSHOT and 8.0.0-SNAPSHOT
This resolves current build error on 8.0.0-SNAPSHOT and makes build successful. 😃 

@sethmlarson . Please review this.


sethmlarson:
We're waiting on fixes in `master` for this. The plan is to deprecate `week_year` and only use `weekyear` but `weekyear` was accidentally removed in master :)

See:
- https://github.com/elastic/elasticsearch/pull/60707
- https://github.com/elastic/elasticsearch/pull/60855

V1NAY8:
Oh! fine. 😃  thanks for the update. Please close this PR

Issue #172, opened 2020-04-10
Implement DataFrame.groupby()


sethmlarson:
See [`pandas.DataFrame.groupby()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html)

Implement either as a transform or a composite aggregation.

stevedodson:
Related to https://github.com/elastic/eland/issues/169

V1NAY8:
@sethmlarson I will try to implement this.
- Is it mirroring pandas.DataFrame.grouby()
- What are the main features that have to be integrated initially?
- Please give me some inputs 😃

sethmlarson:
Just a warning that this change would be a lot of work probably. I'd break it down into the following steps:

- Switch our default aggregation method to use [`composite` agg](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-composite-aggregation.html) so we can paginate aggregation results by default
- Create a `GroupBy` object which holds on to a copy of the QueryCompiler with all the aggregation methods and the fields being grouped on. When calling an agg from this object should add and unpack the bucketing aggs
- Add a `.groupby()` method to `DataFrame` that returns the `GroupBy` object configured properly

I'd submit the first step as one PR and maybe can combine the second two into one PR.

V1NAY8:
Looks good and challenging. Ill get started 😃

V1NAY8:
@sethmlarson For the first step: I understood the concept of composite agg and pagination. 
- You mentioned **switch our default aggregation method**, does this mean edit the _metric_aggs method or create a new method?
    - Because if we change the existing method, composite agg returns buckets each unique value from one column and computes agg for that bucket! We want agg for entire column, we cant compute them from individual buckets 😐 .
    - This returns results for first time and use after_key to return the next results.
    - Should we default keep size as 100?
    - Can you give me one example of the query and how we perform groupby at elasticsearch level using composite aggs?
```bash
curl -X GET "localhost:9200/ecommerce/_search?pretty" -H 'Content-Type: application/json' -d'
{
  "size": 0,
  "aggs": {
        "my_buckets":{
            "composite":{
                "size":2,
                "sources":[
                        {"total_quantity": { "terms": { "field": "total_quantity" } } },
                        {"taxful_total_price": { "terms": { "field": "taxful_total_price" } } }
                  ]
              },
                    "aggregations": {
                        "the_avg": {
                        "avg": { "field": "taxful_total_price" }
[...]

PR #278, opened 2020-10-04
Add groupby to dataframe

V1NAY8:
PR on #172

@sethmlarson  Can you please review these when free? Still lot of work has to be done. 😕 
I implemented groupby to dataframe using composite aggreagation
- Only implemented single aggs and aggregation
- TODO Count, grouped (`ed_flights.groupby("Cancelled")`), dropna

- One more suggestion needed
For the following query eland is taking 7.93 seconds, whereas pandas takes 1.56 seconds 😨 
image

If you can review the composite aggregation logic, and give some inputs on dropna, grouped, count (how to query ES using filters?). **I will proceed on the logic in eland.** That's where I was getting stopped.


sethmlarson (changes requested):
Here are some early comments, I need to pull this PR locally and give it a spin to see what the requests and responses look like :)

eland/operations.py
```diff
@@ -464,6 +404,324 @@ def _hist_aggs(self, query_compiler, num_bins):
...
+        if is_agg:
+            # Convert header columns to MultiIndex
+            agg_df.columns = pd.MultiIndex.from_product([headers, pd_aggs])
+
+        return agg_df
+
+    def _groupby_aggs(
+        self,
+        query_compiler: "QueryCompiler",
+        by: Union[str, List[str]],
```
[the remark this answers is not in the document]
V1NAY8: Yes, at this point I just implemented a list of aggs and single-agg
There are few more options supported in pandas such as naming, different agg for different column ...
Once, the base functionality of groupby is optimized and structured. Ill do them anyway :)

eland/operations.py
```diff
@@ -185,10 +187,29 @@ def aggs(self, query_compiler, pd_aggs, numeric_only=None) -> pd.DataFrame:
     def _metric_aggs(
```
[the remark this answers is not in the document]
V1NAY8: Yay!

V1NAY8:
@sethmlarson 
Just a quick one,
If you comment out the following in `operations.py` at line 701
```python
if b.is_timestamp:
    agg_value = elasticsearch_date_to_pandas_date(
        agg_value, b.es_date_format
    )
```
There will be an improvement of at least 2 seconds 😲 
Then output for the above mentioned query would be:
image

I guess we have to do something in this:
https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html

sethmlarson (changes requested):
Some more comments for you!

The `timestamp` field should be the same as what it'd be if it wasn't in the index so if it's a datetime it should still be a datetime. This will be necessary when we go to implement `resample()`.

eland/field_mappings.py
```diff
@@ -697,14 +699,48 @@ def numeric_source_fields(self):
...
+        aggregatable_fields: List[Field] = []
+        for index_name, row in self._mappings_capabilities.iterrows():
+            row = row.to_dict()
+            row["index"] = index_name
+            if index_name not in by:
+                aggregatable_fields.append(Field(**row))
+            else:
+                # Maintain groupby order as given input
+                groupby_fields[by.index(index_name)] = Field(**row)
+
```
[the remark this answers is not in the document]
V1NAY8: As you mentioned above, will turn it into a dictionary. Hence this will not be required

V1NAY8:
Added tests, Could you please check the performance of groupby with different examples? If possible. Ill optimize if required.
Please add hacktoberfest-accepted to this.

eland/dataframe.py
```diff
@@ -1430,6 +1431,71 @@ def aggregate(
...
+        by: List[str]
+
+        Returns
+        -------
+        GroupByDataFrame
+
+        Examples
+        --------
+        >>> ed_flights = ed.DataFrame('localhost', 'flights', columns=["AvgTicketPrice", "Cancelled", "dayOfWeek", "timestamp", "DestCountry"])
+        >>> ed_flights.groupby(["DestCountry", "Cancelled"]).agg(["min", "max"], True)
```
sethmlarson, line 1451: Use keyword for `numeric_only=True` here for better readability.

eland/dataframe.py
```diff
@@ -1430,6 +1431,71 @@ def aggregate(
 
     hist = gfx.ed_hist_frame
 
+    def groupby(
+        self, by: Union[List[str], Dict[str, str], None] = None, dropna: bool = True
+    ) -> "GroupByDataFrame":
```
sethmlarson, line 1436: Why is `Dict[str, str]` a type allowed by `by`?
Also since `by` is a required parameter we don't have to initiate it with `= None` do we? Or does pandas do this?

eland/dataframe.py
```diff
@@ -1430,6 +1431,71 @@ def aggregate(
...
+        AT          False         100.020531  1181.835815       0.0  6.0
+        ...                              ...          ...       ...  ...
+        TR          True          307.915649   307.915649       0.0  0.0
+        US          False         100.145966  1199.729004       0.0  6.0
+                    True          102.153069  1192.429932       0.0  6.0
+        ZA          False         102.002663  1196.186157       0.0  6.0
+                    True          121.280296  1175.709961       0.0  6.0
+        <BLANKLINE>
+        [63 rows x 4 columns]
+        >>> ed_flights.groupby(["DestCountry", "Cancelled"]).mean(True)
```
sethmlarson, line 1468: `numeric_only=...` here too, should we change this example to use `numeric_only=False`?

eland/field_mappings.py
```diff
@@ -697,14 +698,49 @@ def numeric_source_fields(self):
...
+        by:
+            a list of groupby fields
+
+        Returns
+        -------
+            a Tuple consisting of a list of field mappings for groupby and non-groupby fields
+
+        """
+        groupby_fields: Dict[str, Field] = {}
+        # groupby_fields: Union[List[Field], List[None]] = [None] * len(by)
```
sethmlarson, line 731: You can remove this comment now that the above is working :)

eland/dataframe.py
```diff
@@ -1430,6 +1431,71 @@ def aggregate(
...
+        US          False          598.063146   2.752014
+                    True           579.799066   2.767068
+        ZA          False          636.998605   2.738589
+                    True           677.794078   2.928571
+        <BLANKLINE>
+        [63 rows x 2 columns]
+        """
+        if by is None:
+            raise TypeError("by parameter should be specified to groupby")
+        if isinstance(by, str):
```
sethmlarson, line 1487: Can we move the `KeyErrors` after the normalization has occurred to only be `list`, then include every column that's not in the DataFrame within the `KeyError` message.

sethmlarson (changes requested):
Pulling this locally to take a closer look again! :) Left you a collection of comments. Also a general comment, would be good to start working on more test cases, especially to cover errors.

I need to add `pytest-cov` so we can see how much test coverage we have :)

sethmlarson:
jenkins test this please

eland/dataframe.py
V1NAY8: Yes

eland/field_mappings.py
V1NAY8: 😶 Sorry forgot!

eland/dataframe.py
V1NAY8: So, One example for agg, single-agg for `numeric_only=True` , Should add examples for `numeric_only=False`

eland/dataframe.py
V1NAY8: Ah! I wanted to implement this: 
`ed_df.groupby("Cancelled").agg({"DestCountry": "mean", "AvgTicketPrice": "min"},True)`
But couldn't. Should remove that.
Maybe Ill do another PR for remaining things to be done in groupby

eland/dataframe.py
V1NAY8: sure

eland/groupby.py
```diff
@@ -0,0 +1,173 @@
...
+        List of columns to groupby
+    query_compiler:
+        Query compiler object
+    dropna:
+        default is true, drop None/NaT/NaN values while grouping
+    """
+
+    def __init__(
+        self,
+        by: Union[str, List[str], None] = None,
```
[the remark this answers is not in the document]
V1NAY8: by Need not be none. Will remove it. We should normalize it to List[str].

eland/groupby.py
```diff
@@ -0,0 +1,173 @@
...
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+from abc import ABC
+from typing import List, Optional, Union
+from eland.query_compiler import QueryCompiler
+
+
+class GroupBy(ABC):
```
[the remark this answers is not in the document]
V1NAY8: I intended to add aggregate as an abstract method, but I don't think its necessary, Looks better if we don't inherit ABC. :)

V1NAY8:
- Code Coverage Bot will be good :)
- I think we also need to add a Pull request template that has checks in the Contributing.md doc, 
- Issue template when raised for issues.

eland/field_mappings.py
sethmlarson: No problem, happens to me daily :)

V1NAY8:
- Made groupby strictly typed
- added more tests (I hope this should cover implemented functionality)

sethmlarson:
jenkins test this please

V1NAY8:
Yay! Builds successful 😎

sethmlarson (approved):
This looks good to me! Will do some tweaking and documenting now that the core functionality is there. Thanks for working so hard on this :)

PR #292, opened 2020-10-19
Implement `DataFrameGroupBy.count()`

### LINKED ISSUE #289: Implement DataFrameGroupBy.count()

V1NAY8:
Closes #289 

Implemented
- `ed_df.groupby([...]).count()`
- `ed_df.groupby([...]).agg(['count'])`
- `ed_df.groupby([...]).agg(['max', 'min', 'count', 'mean'])` maintains order
- Fixed bug where following raises exception `ed_df.agg['count']`
- Added tests for all the above functionality
- Add `hacktoberfest-accepted` label

@sethmlarson Please Review :)

### DIFF
--- a/eland/dataframe.py
+++ b/eland/dataframe.py
@@ -1474,6 +1474,7 @@ class DataFrame(NDFrame):
                     True          121.280296  1175.709961       0.0  6.0
         <BLANKLINE>
         [63 rows x 4 columns]
+
         >>> ed_flights.groupby(["DestCountry", "Cancelled"]).mean(numeric_only=True) # doctest: +NORMALIZE_WHITESPACE
                                AvgTicketPrice  dayOfWeek
         DestCountry Cancelled
@@ -1490,6 +1491,23 @@ class DataFrame(NDFrame):
                     True           677.794078   2.928571
         <BLANKLINE>
         [63 rows x 2 columns]
+
+        >>> ed_flights.groupby(["DestCountry", "Cancelled"]).min(numeric_only=False) # doctest: +NORMALIZE_WHITESPACE
+                               AvgTicketPrice  dayOfWeek           timestamp
+        DestCountry Cancelled
+        AE          False          110.799911          0 2018-01-01 19:31:30
+                    True           132.443756          0 2018-01-06 13:03:25
+        AR          False          125.589394          0 2018-01-01 01:30:47
+                    True           251.389603          0 2018-01-01 02:13:17
+        AT          False          100.020531          0 2018-01-01 05:24:19
+        ...                               ...        ...                 ...
+        TR          True           307.915649          0 2018-01-08 04:35:10
+        US          False          100.145966          0 2018-01-01 00:06:27
+                    True           102.153069          0 2018-01-01 09:02:36
+        ZA          False          102.002663          0 2018-01-01 06:44:44
+                    True           121.280296          0 2018-01-04 00:37:01
+        <BLANKLINE>
+        [63 rows x 3 columns]
         """
         if by is None:
             raise ValueError("by parameter should be specified to groupby")
--- a/eland/field_mappings.py
+++ b/eland/field_mappings.py
@@ -100,14 +100,12 @@ class Field(NamedTuple):
             elif es_agg[0] == "percentiles":
                 es_agg = "percentiles"
 
-        if es_agg == "value_count":
-            return True
-        # Cardinality works for all types
-        # Numerics and bools work for all aggs
         # Except "median_absolute_deviation" which doesn't support bool
         if es_agg == "median_absolute_deviation" and self.is_bool:
             return False
-        if es_agg == "cardinality" or self.is_numeric or self.is_bool:
+        # Cardinality and Count work for all types
+        # Numerics and bools work for all aggs
+        if es_agg in ("cardinality", "value_count") or self.is_numeric or self.is_bool:
             return True
         # Timestamps also work for 'min', 'max' and 'avg'
         if es_agg in {"min", "max", "avg", "percentiles"} and self.is_timestamp:
--- a/eland/field_mappings.py
+++ b/eland/field_mappings.py
@@ -100,6 +100,8 @@ class Field(NamedTuple):
             elif es_agg[0] == "percentiles":
                 es_agg = "percentiles"
 
+        if es_agg == "value_count":
+            return True
         # Cardinality works for all types
         # Numerics and bools work for all aggs
         # Except "median_absolute_deviation" which doesn't support bool
@@ -730,7 +732,6 @@ class FieldMappings:
 
         """
         groupby_fields: Dict[str, Field] = {}
-        # groupby_fields: Union[List[Field], List[None]] = [None] * len(by)
         aggregatable_fields: List[Field] = []
         for column, row in self._mappings_capabilities.iterrows():
             row = row.to_dict()
--- a/eland/groupby.py
+++ b/eland/groupby.py
@@ -158,14 +158,19 @@ class GroupByDataFrame(GroupBy):
             A Pandas DataFrame
 
         """
+        # Controls whether a MultiIndex is used for the
+        # columns of the result DataFrame.
+        is_dataframe_agg = True
         if isinstance(func, str):
             func = [func]
+            is_dataframe_agg = False
+
         return self._query_compiler.aggs_groupby(
             by=self._by,
             pd_aggs=func,
             dropna=self._dropna,
             numeric_only=numeric_only,
-            is_dataframe_agg=True,
+            is_dataframe_agg=is_dataframe_agg,
         )
 
     agg = aggregate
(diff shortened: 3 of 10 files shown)


sethmlarson:
Jenkins test this please

eland/field_mappings.py
```diff
@@ -100,6 +100,8 @@ def is_es_agg_compatible(self, es_agg) -> bool:
             elif es_agg[0] == "percentiles":
                 es_agg = "percentiles"
 
+        if es_agg == "value_count":
```
sethmlarson, line 103: Can move this to `es_agg == "cardinality"` -> `es_agg in ("cardinality", "value_count")`

Also should update the comment "Cardinality works for all types" -> "Cardinality and Count work for all types"

eland/field_mappings.py
```diff
@@ -730,7 +732,6 @@ def groupby_source_fields(self, by: List[str]) -> Tuple[List[Field], List[Field]
 
         """
         groupby_fields: Dict[str, Field] = {}
-        # groupby_fields: Union[List[Field], List[None]] = [None] * len(by)
```
sethmlarson, line 733: :+1:

eland/operations.py
```diff
@@ -584,9 +584,58 @@ def aggs_groupby(
         if is_dataframe_agg:
             # Convert header columns to MultiIndex
             agg_df.columns = pd.MultiIndex.from_product([headers, pd_aggs])
+        else:
+            # Convert header columns to Index
+            agg_df.columns = pd.Index(headers)
```
sethmlarson, line 589: Can we add a testcase that verifies this behavior?

eland/operations.py
```diff
@@ -516,7 +516,7 @@ def _unpack_metric_aggs(
                     agg_value = int(agg_value)
 
                 # If this is a non-null timestamp field convert to a pd.Timestamp()
-                elif field.is_timestamp:
+                elif field.is_timestamp and pd_agg != "count":
```
sethmlarson, line 519: Can change the above `elif` to be `elif pd_agg in ("nunique", "count")`

eland/operations.py
```diff
@@ -640,8 +689,19 @@ def _groupby_aggs(
 
         body = Query(query_params.query)
 
+        # To return for creating multi-index on columns
+        headers = [field.column for field in agg_fields]
+
+        pd_aggs_copy = pd_aggs.copy()
```
sethmlarson, line 695: Instead of modifying `pd_aggs` and skipping `unpack` w/ count can we instead move the `doc_count` value into the aggregation within `buckets_generator()` (maybe into `_groupby_bucket_count` to ensure we don't conflict with other values?) and then we can use the normal flow of unpacking aggregations?

Let me know if this isn't the case.

eland/query.py
```diff
@@ -195,8 +195,8 @@ def composite_agg_start(
 
         Parameters
         ----------
-        size: int
-            Pagination size.
+        size: int or None
+            Use composite aggregation with pagination if size is Not None
```
sethmlarson, line 199: nit: lowercase `not`, also when is `size=None`?

eland/groupby.py
```diff
@@ -164,3 +169,14 @@ def aggregate(self, func: List[str], numeric_only: bool = False) -> "pd.DataFram
...
+    def count(self) -> "pd.DataFrame":
+        """
+        Used to groupby and count
+
+        Returns
+        -------
+            A Pandas DataFrame
+
+        """
+        return self._query_compiler.count_groupby(by=self._by, is_dataframe_agg=False)
```
sethmlarson, line 182: Instead of adding a new function `count_groupby()` can we instead use `aggs_groupby()` since it seems that `count_groupby()` simply calls that function?

eland/tests/dataframe/test_groupby_pytest.py
```diff
@@ -154,3 +154,17 @@ def test_groupby_different_dtypes(self, by):
     def test_groupby_dropna(self):
         # TODO Add tests once dropna is implemeted
         pass
+
+    def test_groupByDataFrame_count(self):
```
sethmlarson, line 158: nit: use snake case not camelcase: `test_groupby_dataframe_count()`

sethmlarson (changes requested):
Woo, this is great :) Some comments for you!

eland/tests/dataframe/test_groupby_pytest.py
V1NAY8: Oops 😐

eland/groupby.py
V1NAY8: Yeah!, I totally missed that.

eland/query.py
V1NAY8: - size=None is for checking, If we want pagination or not.
- If size is given, pagination can be performed.

eland/operations.py
V1NAY8: Yep

eland/field_mappings.py
V1NAY8: Yep

eland/dataframe.py
```diff
@@ -1490,6 +1491,23 @@ def groupby(
                     True           677.794078   2.928571
         <BLANKLINE>
         [63 rows x 2 columns]
+
+        >>> ed_flights.groupby(["DestCountry", "Cancelled"]).min(numeric_only=False) # doctest: +NORMALIZE_WHITESPACE
```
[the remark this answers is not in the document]
V1NAY8: I forgot to add this old test case last time

eland/operations.py
V1NAY8: - My intention was not to pass count to `_unpack`, so that we would use `doc_count` especially for groupby instead of using `count` of `metric_aggs`. 
- The original pd_aggs is used at end for construction of dictionary with all aggs (along with count)

This was needed because, we had common method for both `metric_aggs` and `groupby_aggs` . To eliminate the copy , we can introduce a new parameter `is_Groupby` to skip if 'count' agg is passed to it.

Do you have any better Idea ?

eland/operations.py
V1NAY8: Yes :)

eland/dataframe.py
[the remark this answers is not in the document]
V1NAY8: I also thought, we need to add documentation for groupby.py file. And add each doctest for each groupby ??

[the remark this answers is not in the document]
sethmlarson: It'd be nice to have a doctest for each function. Maybe I can work on this if you don't want to, it'll just be a lot of copy-paste :P I'll create an issue for it.

eland/query.py
V1NAY8: Ok, Sorry size=None doesn't make sense

eland/operations.py
V1NAY8: Usually doc_count is returned by all bucket by default for all composite aggregation
E.g.: 
```json
"buckets" : [
        {
          "key" : {
            "groupby_Cancelled" : false
          },
          "doc_count" : 11381
        },
        {
          "key" : {
            "groupby_Cancelled" : true
          },
          "doc_count" : 1678
        }
      ]
```
The logic that I have chosen 
is to consider doc_count if `count` is given in pd_agg and not consider finding aggregation on `count` again

- To be honest, I wanted to achieve all in single query

sethmlarson: I'm going to try pulling the PR locally and playing around.

eland/query.py
```diff
@@ -215,10 +215,13 @@ def composite_agg_start(
...
         self._composite_aggs.clear()
 
-        aggs = {
-            "composite": {"size": size, "sources": sources},
-            "aggregations": self._aggs.copy(),
+        aggs: Dict[str, Dict[str, Any]] = {
+            "composite": {"size": size, "sources": sources}
         }
+
+        if self._aggs:
```
sethmlarson, line 222: Nice, no guarantee there'll actually be `aggs` if we're simply calling `count()`

sethmlarson (commented):
Everything except the core functionality looks good, I'm going to pull locally and try things out to better review.

sethmlarson:
Pulled it locally and I think I found a way to work around the `count` issue by adding a fake result to each bucket that gets unpacked properly in `_unpack_metric_aggs()`. I've attached it, you can apply it via: `git apply eland-groupby-count.txt`

[eland-groupby-count.txt](https://github.com/elastic/eland/files/5425706/eland-groupby-count.txt)

(It's not a `.patch` file because apparently GitHub doesn't allow embedding those to comments?)

V1NAY8:
Wait.. What ?? We could do it that simple, Without additional logic for maintaining order`. That was brilliant. 🤯 
I guess now everything is optimized..

sethmlarson (approved):
This looks good to me, thanks for prettying it up! :)

sethmlarson:
jenkins test this please

V1NAY8:
Builds looking good :)

sethmlarson:
Yeah went to grab coffee! :coffee: Thank you for this PR :)

PR #298, opened 2020-10-26
Add nbval to CI and notebook examples

### LINKED ISSUE #88: Add nbval to CI for jupyter notebook tests
More information about nbval here: 
https://pypi.org/project/nbval/

V1NAY8:
Closes #88 
The following changes are made:
- Modified `noxfile.py` to include `nbval` tests which are executed when CI is run.
- Added some notebook tests which are 
    - `# doctest: +SKIP` tests in tests folder
    -  A demo notebook of all basic functionality
    -  ETL notebook which consists of tests for `show_progress` parameter
    -  Metrics notebook which consists of `median`, `mad`  => `numeric_only` tests

@sethmlarson Please review and let me know if any tests are to be added or If existing ones are not necessary. 
- License headers are not included for `.ipynb` or `.csv` files, Is that Acceptable ?


noxfile.py
```diff
@@ -100,7 +100,11 @@ def lint(session):
 def test(session):
     session.install("-r", "requirements-dev.txt")
     session.run("python", "-m", "eland.tests.setup_tests")
+    session.run("python", "setup.py", "install")
     session.run("pytest", "--doctest-modules", *(session.posargs or ("eland/",)))
+    session.run(
```
sethmlarson, line 105: Let's add a check that sees if `session.posargs` aren't empty and then only if they are empty do we run the `--nbval` or `eland/tests/ml` below, that way if you're targeting a single test with `nox -rs test-3.8 -- eland/tests/test_...` then ONLY those tests are executed.

Does this make sense?

noxfile.py
```diff
@@ -100,7 +100,11 @@ def lint(session):
 def test(session):
     session.install("-r", "requirements-dev.txt")
     session.run("python", "-m", "eland.tests.setup_tests")
+    session.run("python", "setup.py", "install")
```
sethmlarson, line 103: Prefer `session.install(".")` over `setup.py install`

sethmlarson (changes requested):
Thanks for this! Two comments for you :)

sethmlarson:
jenkins test this please

sethmlarson:
@V1NAY8 Also it's fine that license headers aren't in non-.py files.

noxfile.py
V1NAY8: I only wanted to execute the notebook tests. So, Ill remove the `session.posargs`
Sorry, I really don't know what `session.posargs` does.

sethmlarson: Oh no problem, `session.posargs` is essentially a tuple of string arguments that can be passed in via `nox -s <session> -- <posargs>`

for example if I were to run `nox -s test-3.8 -- tests/a tests/b` then `session.posargs` would be `("tests/a", "tests/b")`

The feature isn't super-well documented on the Nox side unfortunately but very useful for test suites especially.

V1NAY8: Yeah  thanks

V1NAY8:
Also, the build on master failed. after merging my last PR :(

Trying to fix it, 
The issue is at 
`pytest --doctest-modules .\eland\tests\dataframe\test_utils_pytest.py::TestDataFrameUtils::test_es_type_override_error`
There is a regex error coming on this `column(s)`
But if `column/s` is kept then it is resolved.

I better change it to `columns` ?

sethmlarson:
No problem, change the expected message to `column(s)` and we'll run the CI again.

V1NAY8:
The expected message is also column(s) and in field mappings is also column(s). But they aren't matching each other.

sethmlarson:
Oh, are we using regular expressions? Instead you should use an exact match:

```python
with pytest.raises(ValueError) as e:
    ...
assert str(e.value) == "...column(s)..."
```

It's failing because `(` and `)` have special meaning to regexes

V1NAY8:
Ask jenkins to test

V1NAY8:
BTW, when is the next eland release ?

sethmlarson:
jenkins test this please

sethmlarson:
Next release will be `v7.10.0a1` which we could probably release soon. Lots of work excited features! :)

V1NAY8:
Builds are good :)

sethmlarson (approved):
LGTM :rocket:

PR #299, opened 2020-10-26
Add tqdm to pandas_to_eland()

### LINKED ISSUE #294: Add TQDM progress bar for Jupyter Notebook pandas_to_eland() calls
- https://github.com/tqdm/tqdm#ipythonjupyter-integration
- Units are rows/s
- Maybe hide progress bar on completion? Need to test it out first.
- Add an option `show_progressbar` which defaults to `None`
  - If `None` show a progress bar only if we detect Jupyter Notebook
  - If `False` don't show a progress bar
  - If `True` show a progress bar only if we detect Jupyter Notebook (for now)

V1NAY8:
Closes #294 

- Added tqdm functionality for `pandas_to_eland()`

- If we want the progress bar to disappear we will need to just add `leave=False` to tqdm context manager at line 234

@sethmlarson

Please review if the implementation is correct, I will add notebook tests for this once review is done. 😃


V1NAY8:
Added tests, updated `requirements*.txt`

eland/etl.py
```diff
@@ -209,35 +216,55 @@ def pandas_to_eland(
...
+                yield action
 
     # parallel_bulk is lazy generator so use deque to consume them immediately
     # maxlen = 0 because don't need results of parallel_bulk
     deque(
         parallel_bulk(
             client=es_client,
             actions=action_generator(
-                pd_df, es_dropna, use_pandas_index_for_es_ids, es_dest_index
+                pd_df,
```
sethmlarson, line 263: nit: can you use keyword arguments here?

eland/etl.py
```diff
@@ -22,7 +22,9 @@
 import pandas as pd  # type: ignore
 from elasticsearch import Elasticsearch
 from elasticsearch.helpers import parallel_bulk
+from pandas.core.config_init import is_terminal  # type: ignore
```
sethmlarson, line 25: I'm not sure if `is_terminal` detects only Jupyter Notebook, from the function itself it looks like it mentions IPython which can be used outside of a Jupyter Notebook: https://github.com/pandas-dev/pandas/blob/master/pandas/core/config_init.py#L323

requirements.txt
```diff
@@ -1,3 +1,4 @@
 elasticsearch>=7.7
 pandas>=1
 matplotlib
+tqdm
```
sethmlarson, line 4: You'll have to add `tqdm` to `setup.py` `install_requires` as well

sethmlarson (changes requested):
Thanks, few comments for you

eland/etl.py
V1NAY8: Should I do this way?
```python
        try:
            get_ipython()
        except NameError:
            pass
```
`get_ipython()` works in jupyter notebook whereas it fails in terminal.

V1NAY8:
Yep Seth, Made changes, 
Also in `setup.py` `install-requires` has `elasticsearch>=7.7`, but in our jenkins matrix we also test for 7.6.
Should I remove it?

sethmlarson:
I'll handle the elasticsearch-py stuff, there's a subtle change that needs to be made regarding which elasticsearch-py version to install

V1NAY8:
Ask jenkins to test this please

sethmlarson:
jenkins test this please

V1NAY8:
Also I took reference for ipython check from 
https://github.com/ipython/ipython/issues/9732#issuecomment-231592556

V1NAY8:
@sethmlarson Anything else required on this change?

PR #301, opened 2020-10-26
Add support for es_match() to DataFrame and Series

### LINKED ISSUE #282: Support Elasticsearch full-text queries
Since Pandas doesn't have many full-text search APIs we can prefix these with `es_` and take Elasticsearch options:

https://www.elastic.co/guide/en/elasticsearch/reference/current/full-text-queries.html

I'm thinking `Series.es_match()` which returns a filter to be used like so:

`df[df.address.es_match("University Ave")]`

and potentially a `DataFrame.es_match()` which allows passing one or more fields (`match`/`multi_match`) for a given query defaulting to all text fields.

sethmlarson:
Closes #282

### DIFF
--- a/eland/series.py
+++ b/eland/series.py
@@ -55,6 +55,7 @@ from eland.filter import (
     LessEqual,
     NotFilter,
     NotNull,
+    QueryFilter,
     ScriptFilter,
 )
 from eland.ndframe import NDFrame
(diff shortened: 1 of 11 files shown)


eland/series.py
```diff
@@ -620,6 +621,74 @@ def filter(
...
+            Boolean filter to be combined with other filters and
+            then passed to DataFrame[...].
+
+        Examples
+        --------
+        >>> df = ed.DataFrame("localhost:9200", "ecommerce")
+        >>> df[
+        ...     df.category.es_match("Men's")
+        ...     & (df.taxful_total_price > 200.0)
+        ... ].filter([
```
V1NAY8, line 669: So, just change the order
```python
  ... ].filter([
        ...     "category",
        ...     "taxful_total_price"
        ... ]).head(5)
```
And rebase, force-push. Builds will be good :)

V1NAY8 (changes requested):

V1NAY8:
I tried adding commits, but couldn't.

eland/series.py
sethmlarson: Thanks, fixed this.

sethmlarson:
@V1NAY8 Yeah I think only users with write privileges can push to other's branches.

V1NAY8:
Ah! I get it, I usually do check the `Allow edits and access to secrets by maintainers` checkbox. So, you could push commit on top of my PR 😁

PR #302, opened 2020-10-27
es_type_overrides should error when receiving non-existing entries

### LINKED ISSUE #280: es_type_overrides should error when receiving non-existing entries

V1NAY8:
Closes #280 

@sethmlarson Please review 😄

### DIFF
--- a/eland/field_mappings.py
+++ b/eland/field_mappings.py
@@ -42,8 +42,6 @@ from pandas.core.dtypes.inference import is_list_like
 if TYPE_CHECKING:
     from elasticsearch import Elasticsearch
 
-    from eland import DataFrame
-
 
 ES_FLOAT_TYPES: Set[str] = {"double", "float", "half_float", "scaled_float"}
 ES_INTEGER_TYPES: Set[str] = {"long", "integer", "short", "byte"}
@@ -463,7 +461,7 @@ class FieldMappings:
         return cls.ES_DTYPE_TO_PD_DTYPE.get(es_dtype, "object")
 
     @staticmethod
-    def _pd_dtype_to_es_dtype(pd_dtype):
+    def _pd_dtype_to_es_dtype(pd_dtype) -> str:
         """
         Mapping pandas dtypes to Elasticsearch dtype
         --------------------------------------------
@@ -479,7 +477,7 @@ class FieldMappings:
         category NA NA Finite list of text values
         ```
         """
-        es_dtype = None
+        es_dtype: str = None
 
         # Map all to 64-bit - TODO map to specifics: int32 -> int etc.
         if is_float_dtype(pd_dtype):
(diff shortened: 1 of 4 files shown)


sethmlarson:
Jenkins test this please

V1NAY8:
Builds are good :)

eland/field_mappings.py
```diff
@@ -537,7 +535,17 @@ def _generate_es_mappings(
         }
         """
 
-        mapping_props = {}
+        mapping_props: Dict[str, Any] = {}
+        es_dtype: str
```
sethmlarson, line 539: Is this type hint necessary? If it is let's move it to the top of the function with whitespace before the `mapping_props = {}` so we don't confuse it for an initialization.

eland/field_mappings.py
```diff
@@ -479,7 +477,7 @@ def _pd_dtype_to_es_dtype(pd_dtype):
         category NA NA Finite list of text values
         ```
         """
-        es_dtype = None
+        es_dtype: str = None
```
sethmlarson, line 480: This function can return `None` so should be `Optional[str]` both here and return type?

eland/field_mappings.py
```diff
@@ -537,7 +535,17 @@ def _generate_es_mappings(
...
-        mapping_props = {}
+        mapping_props: Dict[str, Any] = {}
+        es_dtype: str
+        if es_type_overrides is not None:
+            non_existing_columns: List[str] = [
+                key for key in es_type_overrides.keys() if key not in dataframe.columns
+            ]
+            if non_existing_columns:
+                raise KeyError(
+                    f"{repr(non_existing_columns)[1:-1]} column/s not in given dataframe"
```
sethmlarson, line 546: nit: `column(s)` instead of `column/s`

sethmlarson (changes requested):
Few small comments then looks good!

eland/field_mappings.py
V1NAY8: Just added, Anyway will move it as suggested

eland/field_mappings.py
V1NAY8: Yep, Missed in plain sight 👍

V1NAY8:
CI workflows are failing because, we are using elasticsearch 
Hence we need to add `python3.8 -m pip install nox elasticsearch` in yml

You want me to add it?

sethmlarson:
Instead of installing `elasticsearch` can you move the `import elasticsearch` from the global scope to inside the `try-except` for `es_active`?

V1NAY8:
Hmm. Now pandoc is not installed 🤯

sethmlarson:
ah, likely need a `apt-get install pandoc` within the github action?

sethmlarson:
@V1NAY8 I pushed to your branch, recommend you do a pull.

V1NAY8:
Yes, pulled it, Now the CI are successful .

sethmlarson (approved):
Woo, thank you for this! :tada: :partying_face:

PR #303, opened 2020-10-27
Add linting GitHub Actions, fix docs

sethmlarson:
Also waiting for the RTD commit status so I can require it on PRs :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,36 @@
+name: CI
+
+on: [push, pull_request]
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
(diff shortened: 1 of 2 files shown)


V1NAY8:
Just asking: 
- I think we should also do  `nox -s docs` to check if errors are being captured while building docs.
- Any reason why using `nox -s lint` instead of `nox -s format` ?

sethmlarson:
@V1NAY8 

* Was trying to use RTD PR builds for this, however it seems the OAuth app isn't installed in the elastic org so will have to use GHA.
* `nox -s format` modifies code, we would want that to be run before commit. `nox -s lint` only checks that code is proper which is what we want.

V1NAY8:
I also think, to stop Jenkins do the `nox -s lint` and just do `nox -s format`
I guess we also need to remove:
https://github.com/elastic/eland/blob/28951c0ad1a8c3cc8b57035023dca80c059785bc/noxfile.py#L65

PR #307, opened 2020-10-27
Test against Elasticsearch 7.10

sethmlarson:
Stop testing Python 3.7 because it didn't add anything interesting. 3.6 covers for it.


V1NAY8:
I think we also need to add 3.9 here
https://github.com/elastic/eland/blob/95b8d75e372cb099450a4e3bd79936021c95f377/noxfile.py#L98

- Also I tried creating virtualenv locally with 3.9 to check pip installations.
- Seems like `pip install numpy pandas matplotlib scikit-learn` (checked each package individually) with python 3.9 failed. 
- Also tried doing `sudo apt-get install python-dev -y` and all the above, still we are getting error. 
🤔

sethmlarson:
Yeah looks like those dependencies don't have wheels for 3.9 yet. Oh well, they will one day :)

PR #323, opened 2020-11-12
Add mode to dataframe and series

### LINKED ISSUE #215: Implement [DataFrame, Series].mode aggregation via top terms
See the TODO comment in `operations.py`:

V1NAY8:
Closes #215 

- Added mode to dataframe and series
- Added tests, documentation
- Currently mode isn't supported by `pandas groupby` too, Hence added a `NotImplementedError`
- To satisfy `mypy` I had to add type hints to some other methods too.

- Implemented es_size parameter instead of dumping the entire cluster if we have multiple mode values.

- Currently dropna=True is only supported because I have a query on `missing` parameter of terms aggregation
[Reference](https://github.com/elastic/eland/issues/215#issuecomment-725407371)

@sethmlarson Please review 😃

### DIFF
--- a/docs/sphinx/reference/dataframe.rst
+++ b/docs/sphinx/reference/dataframe.rst
@@ -89,6 +89,7 @@ Computations / Descriptive Stats
    DataFrame.var
    DataFrame.sum
    DataFrame.nunique
+   DataFrame.mode
 
 Reindexing / Selection / Label Manipulation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/eland/field_mappings.py
+++ b/eland/field_mappings.py
@@ -102,15 +102,16 @@ class Field(NamedTuple):
         # Except "median_absolute_deviation" which doesn't support bool
         if es_agg == "median_absolute_deviation" and self.is_bool:
             return False
-        # Cardinality and Count work for all types
+        # Cardinality, Count and mode work for all types
         # Numerics and bools work for all aggs
-        if es_agg in ("cardinality", "value_count") or self.is_numeric or self.is_bool:
+        if (
+            es_agg in ("cardinality", "value_count", "mode")
+            or self.is_numeric
+            or self.is_bool
+        ):
             return True
         # Timestamps also work for 'min', 'max' and 'avg'
         if es_agg in {"min", "max", "avg", "percentiles"} and self.is_timestamp:
-            return True
-        # All datatypes work for mode
-        if es_agg == "mode":
             return True
         return False
 
--- a/eland/field_mappings.py
+++ b/eland/field_mappings.py
@@ -109,6 +109,9 @@ class Field(NamedTuple):
         # Timestamps also work for 'min', 'max' and 'avg'
         if es_agg in {"min", "max", "avg", "percentiles"} and self.is_timestamp:
             return True
+        # All datatypes work for mode
+        if es_agg == "mode":
+            return True
         return False
 
     @property
--- a/eland/field_mappings.py
+++ b/eland/field_mappings.py
@@ -105,7 +105,7 @@ class Field(NamedTuple):
         # Cardinality, Count and mode work for all types
         # Numerics and bools work for all aggs
         if (
-            es_agg in ("cardinality", "value_count", "mode")
+            es_agg in {"cardinality", "value_count", "mode"}
             or self.is_numeric
             or self.is_bool
         ):
--- a/eland/operations.py
+++ b/eland/operations.py
@@ -571,16 +571,17 @@ class Operations:
                     # include top-terms in the result.
                     if not agg_value:
                         # If the all the documents for a field are empty
-                        agg_value = [np.NaN]
+                        agg_value = [field.nan_value]
                     else:
                         max_doc_count = agg_value[0]["doc_count"]
                         # We need only keys which are equal to max_doc_count
                         # lesser values are ignored
                         agg_value = [
-                            agg_value[i]["key"]
-                            for i in range(len(agg_value))
-                            if agg_value[i]["doc_count"] == max_doc_count
+                            item["key"]
+                            for item in agg_value
+                            if item["doc_count"] == max_doc_count
                         ]
+
                         # Maintain datatype by default because pandas does the same
                         # text are returned as-is
                         if field.is_bool or field.is_numeric:
--- a/eland/operations.py
+++ b/eland/operations.py
@@ -181,7 +181,7 @@ class Operations:
                 dtype = "object"
             return build_pd_series(results, index=results.keys(), dtype=dtype)
 
-    def value_counts(self, query_compiler, es_size):
+    def value_counts(self, query_compiler: "QueryCompiler", es_size: int) -> pd.Series:
         return self._terms_aggs(query_compiler, "terms", es_size)
 
     def hist(self, query_compiler, bins):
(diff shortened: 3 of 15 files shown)


eland/field_mappings.py
```diff
@@ -109,6 +109,9 @@ def is_es_agg_compatible(self, es_agg) -> bool:
         # Timestamps also work for 'min', 'max' and 'avg'
         if es_agg in {"min", "max", "avg", "percentiles"} and self.is_timestamp:
             return True
+        # All datatypes work for mode
+        if es_agg == "mode":
```
sethmlarson, line 113: You can move this into the `es_agg in ("cardinality", ...)` section

eland/operations.py
```diff
@@ -195,12 +195,54 @@ def aggs(self, query_compiler, pd_aggs, numeric_only=None) -> pd.DataFrame:
...
+            pd_aggs=pd_aggs,
+            numeric_only=numeric_only,
+            dropna=dropna,
+            es_mode_size=es_size,
+        )
+
+        pd_dict: Dict[str, Any] = {}
+        row_diff: Optional[int] = None
+
+        if is_dataframe:
```
sethmlarson, line 219: Nice! :clap:

eland/operations.py
```diff
@@ -499,13 +556,42 @@ def _unpack_metric_aggs(
...
                 else:
                     agg_value = response["aggregations"][
                         f"{es_agg}_{field.es_field_name}"
                     ]["value"]
 
+                if isinstance(agg_value, list):
+                    # include top-terms in the result.
+                    if not agg_value:
+                        # If the all the documents for a field are empty
+                        agg_value = [np.NaN]
```
sethmlarson, line 574: Do we need to use `field.nan_value` instead of `np.NaN` here?

eland/operations.py
```diff
@@ -499,13 +556,42 @@ def _unpack_metric_aggs(
...
+                    else:
+                        max_doc_count = agg_value[0]["doc_count"]
+                        # We need only keys which are equal to max_doc_count
+                        # lesser values are ignored
+                        agg_value = [
```
sethmlarson, line 579: Can use this instead of indexing?

```python
agg_value = [
    agg_value_item["key"]
    for item in agg_value
    if agg_value_item["doc_count"] == max_doc_count
]
```

eland/operations.py
```diff
@@ -517,13 +603,22 @@ def _unpack_metric_aggs(
...
+                        ]
+                    else:
+                        agg_value = elasticsearch_date_to_pandas_date(
+                            agg_value, field.es_date_format
+                        )
                 # If numeric_only is False | None then maintain column datatype
                 elif not numeric_only:
                     # we're only converting to bool for lossless aggs like min, max, and median.
-                    if pd_agg in {"max", "min", "median", "sum"}:
+                    if pd_agg in {"max", "min", "median", "sum", "mode"}:
```
sethmlarson, line 621: Nice

eland/tests/dataframe/test_metrics_pytest.py
```diff
@@ -424,3 +424,18 @@ def test_aggs_count(self):
         ed_count = ed_flights.agg(["count"])
 
         assert_frame_equal(pd_count, ed_count)
+
+    @pytest.mark.parametrize("numeric_only", [True, False])
```
sethmlarson, line 428: Can we add a test case that shows that `es_size` works as expected? (ie have a situation where there are more than 1 mode, set es_size>N and see that the dataframe is ==N, then set es_size<N and see that dataframe is es_size still.

In this same test case you can likely test the "fill" mechanic too

sethmlarson (changes requested):
This is really great, here are some comments for you!

eland/operations.py
V1NAY8: Exactly, It serves the purpose

eland/operations.py
V1NAY8: Yes, better than range()

eland/tests/dataframe/test_metrics_pytest.py
V1NAY8: Absolutely

V1NAY8:
If you're happy with the change, I will rebase my branch and resolve merge conflicts :)

sethmlarson:
jenkins test this please

eland/field_mappings.py
```diff
@@ -102,9 +102,13 @@ def is_es_agg_compatible(self, es_agg) -> bool:
         # Except "median_absolute_deviation" which doesn't support bool
         if es_agg == "median_absolute_deviation" and self.is_bool:
             return False
-        # Cardinality and Count work for all types
+        # Cardinality, Count and mode work for all types
         # Numerics and bools work for all aggs
-        if es_agg in ("cardinality", "value_count") or self.is_numeric or self.is_bool:
+        if (
+            es_agg in ("cardinality", "value_count", "mode")
```
sethmlarson, line 108: Change this to a set `{"cardinality", "value_count", "mode"}` instead of tuple now that 3+ elements

sethmlarson (commented):
Sorry for letting this go so long without a final review, I'm pretty sure this is ready but want to run it against the new pandas v1.2.0, can you rebase?

eland/field_mappings.py
V1NAY8: Yep

V1NAY8:
I forgot to include `recursive-include eland` in `manifest.in`. So, I included it. Otherwise the files won't come to release packages.

V1NAY8:
Ask jenkins to test this.

sethmlarson:
jenkins test this please

V1NAY8:
CI is good 🔥

eland/operations.py
```diff
@@ -195,12 +195,54 @@ def aggs(self, query_compiler, pd_aggs, numeric_only=None) -> pd.DataFrame:
...
+        if is_dataframe:
+            # If multiple values of mode is returned for a particular column
+            # find the maximum length and use that to fill dataframe with NaN/NaT
+            rows_len = max([len(value) for value in results.values()])
+            for key, values in results.items():
+                row_diff = rows_len - len(values)
+                # Convert np.ndarray to list
+                values = list(values)
+                if row_diff:
+                    if isinstance(values[0], pd.Timestamp):
```
sethmlarson, line 228: If we call `.mode()` on a dataframe that has no entries for a field but entries in other fields should we look at the es_type on the field instead of assuming if the value is a `pd.Timestamp` then we put `NaT`?

eland/operations.py
```diff
@@ -791,10 +887,15 @@ def _map_pd_aggs_to_es_aggs(pd_aggs):
             elif pd_agg == "median":
                 es_aggs.append(("percentiles", "50.0"))
 
-            # Not implemented
             elif pd_agg == "mode":
-                # We could do this via top term
-                raise NotImplementedError(pd_agg, " not currently implemented")
+                if len(pd_aggs) != 1:
+                    raise NotImplementedError(
+                        "Currently mode is not supported in df.agg[...]. Try df.mode()"
```
sethmlarson, line 893: nit: `df.agg(...)` instead of `df.agg[...]`

tests/dataframe/test_metrics_pytest.py
```diff
@@ -426,3 +426,19 @@ def test_aggs_count(self):
         ed_count = ed_flights.agg(["count"])
 
         assert_frame_equal(pd_count, ed_count)
+
+    @pytest.mark.parametrize("numeric_only", [True, False])
+    @pytest.mark.parametrize("es_size", [2, 20, 100, 5000, 3000])
```
sethmlarson, line 431: Can we add `es_size=1`?

tests/series/test_metrics_pytest.py
```diff
@@ -114,3 +115,25 @@ def test_flights_datetime_median_metric(self):
...
+    def test_ecommerce_mode(self, column):
+        ed_series = self.ed_ecommerce()
+        pd_series = self.pd_ecommerce()
+
+        ed_mode = ed_series[column].mode()
+        pd_mode = pd_series[column].mode()
+
+        assert_series_equal(ed_mode, pd_mode)
+
+    @pytest.mark.parametrize("es_size", [2, 10, 20])
```
sethmlarson, line 131: Can we try `es_size=1` as well?

sethmlarson (changes requested):
This is almost done! Here's a couple more comments:

eland/operations.py
V1NAY8: We should always look at `field.nan_value`. But at this point of code we just return a dictionary that contains results.
I should re-fetch the fields using `fields = query_compiler._mappings.all_source_fields()`

and then if we have a row difference then it should look `values.extend([field.nan_value] * row_diff)`.

This should eliminate checking if it is timestamp or not.

@sethmlarson Shall I make it this way?

sethmlarson: Eh, this is fine for now, however I wonder if what we're seeing here indicates that instead of returning a dictionary of {str: result} for results we should return a dictionary of {Field: result}? Can be done later.

V1NAY8:
I had to make one small change to tests because
Even though the values are same, dtypes are being differed. I am hoping this shouldn't be a thing.
But dtype checks for other es_size are working.
```
>>> pd_testing.mode()[:1] 
  Cancelled  dayOfWeek           timestamp DestCountry FlightNum
0     False          0 2018-01-09 07:54:19          IT   00882F6
>>> ed_testing.mode(es_size=1) 
   Cancelled  dayOfWeek           timestamp DestCountry FlightNum
0      False          0 2018-01-09 07:54:19          IT   00882F6
```
```
assert_frame_equal(ed_testing.mode(es_size=1),pd_testing.mode()[:1])
AssertionError: Attributes of DataFrame.iloc[:, 0] (column name="Cancelled") are different

[left]:  bool
[right]: object
```

docs/sphinx/reference/dataframe.rst
```diff
@@ -89,6 +89,7 @@ Computations / Descriptive Stats
    DataFrame.var
    DataFrame.sum
    DataFrame.nunique
```
sethmlarson, line 91: We're actually missing entries for `DataFrame.mode` and `Series.mode`, need to create a file under `docs/sphinx/reference/api/eland.DataFrame.mode.rst`:

```rst
eland.DataFrame.mode
====================

.. currentmodule:: eland

.. automethod:: DataFrame.mode

```

and the same for `Series.mode`

sethmlarson (changes requested):
Just one more comment, otherwise looks good to me!

sethmlarson:
jenkins test this please

V1NAY8:
Added those files.

sethmlarson (approved):
LGTM, thank you!

PR #329, opened 2020-11-20
Refactor manifest.in

V1NAY8:
Since we don't have `py.typed` in root directory
In `Manifest.in`
Either it should be included as part of `recursive-include eland`
or make it `include eland/py.typed`

I chose the second.

@sethmlarson Need thoughts on this.


sethmlarson:
jenkins test this please

sethmlarson (approved):
LGTM, will merge once CI passes!

PR #335, opened 2020-12-29
Switch to GitHub Actions for matrix CI

sethmlarson:


V1NAY8:
So, are we completely migrating from Jenkins to GHA ?

sethmlarson:
Just testing things out, I think because Pandas v1.2 only supports Python 3.7+ we're actually covered with just Jenkins.

PR #343, opened 2021-03-07
Add implicit-reexport to mypy

V1NAY8:
This PR is to fix pipelines.

The following error was thrown:
```
00:55:26 eland/ml/transformers/__init__.py:67: error: Module 'eland.ml.transformers.xgboost' does not explicitly export attribute 'XGBClassifier'; implicit reexport disabled
00:55:26 eland/ml/transformers/__init__.py:67: error: Module 'eland.ml.transformers.xgboost' does not explicitly export attribute 'XGBRegressor'; implicit reexport disabled
00:55:26 eland/ml/transformers/__init__.py:88: error: Module 'eland.ml.transformers.lightgbm' does not explicitly export attribute 'LGBMClassifier'; implicit reexport disabled
00:55:26 eland/ml/transformers/__init__.py:88: error: Module 'eland.ml.transformers.lightgbm' does not explicitly export attribute 'LGBMRegressor'; implicit reexport disabled.
```
This is because these are implicit exports and mypy is not allowing them. My approach is to enable --implicit-reexports.

According to https://github.com/python/mypy/issues/7042#issue-459406387

mypy should treat items in `__all__`  as exported. Not sure why this is happening.

@sethmlarson Any thoughts? Or Please review this and merge it into master.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -68,12 +68,12 @@ def lint(session):
     session.run("flake8", "--ignore=E501,W503,E402,E712,E203", *SOURCE_FILES)
 
     # TODO: When all files are typed we can change this to .run("mypy", "--strict", "eland/")
-    session.log("mypy --strict eland/")
+    session.log("mypy --implicit-reexport --strict eland/")
     for typed_file in TYPED_FILES:
         if not os.path.isfile(typed_file):
             session.error(f"The file {typed_file!r} couldn't be found")
         popen = subprocess.Popen(
-            f"mypy --strict {typed_file}",
+            f"mypy --implicit-reexport --strict {typed_file}",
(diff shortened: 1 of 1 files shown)


sethmlarson:
Hmm, is there a way we can fix the issue without the flag? I'd prefer to go that direction if possible. This may be as simple as using `from .lightgbm import LGBMClassifier as LGBMClassifier` for all re-exported imports within `transformers/__init__.py`.

stevedodson:
Hopefully resolved with eland/ml/transformers/__init__.py in https://github.com/elastic/eland/pull/346

sethmlarson:
Closed in favor of #346

PR #354, opened 2021-06-08
Remove Warnings because of deprecations

V1NAY8:
So, There are around 184 deprecations because of the following:
- `check_less_precise` is deprecated and hence replaced it with `rtol` Ref: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_frame_equal.html
- `XGBClassifier` has deprecated `use_label_encoder` and we should keep it as `False` Ref: https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBClassifier
- `Numpy` wants us to use `np.int_` instead of `np.int` Ref: https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

### DIFF
--- a/tests/ml/test_ml_model_pytest.py
+++ b/tests/ml/test_ml_model_pytest.py
@@ -263,10 +263,12 @@ class TestMLModel:
             training_data = datasets.make_classification(
                 n_features=5, n_classes=3, n_informative=3
             )
-            classifier = XGBClassifier(booster="gbtree", objective="multi:softmax")
+            classifier = XGBClassifier(
+                booster="gbtree", objective="multi:softmax", use_label_encoder=False
+            )
         else:
             training_data = datasets.make_classification(n_features=5)
-            classifier = XGBClassifier(booster="gbtree")
+            classifier = XGBClassifier(booster="gbtree", use_label_encoder=False)
 
         # Train model
         classifier.fit(training_data[0], training_data[1])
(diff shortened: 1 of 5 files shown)


tests/ml/test_ml_model_pytest.py
```diff
@@ -263,10 +263,12 @@ def test_xgb_classifier(self, compress_model_definition, multi_class):
...
                 n_features=5, n_classes=3, n_informative=3
             )
-            classifier = XGBClassifier(booster="gbtree", objective="multi:softmax")
+            classifier = XGBClassifier(
+                booster="gbtree", objective="multi:softmax", use_label_encoder=False
+            )
         else:
             training_data = datasets.make_classification(n_features=5)
-            classifier = XGBClassifier(booster="gbtree")
+            classifier = XGBClassifier(booster="gbtree", use_label_encoder=False)
```
sethmlarson, line 271: @benwtrent Is this fine for you?

sethmlarson (approved):
Thanks for this!! Except for the XGBoost change I'm confident on these changes. @benwtrent can you take a quick peek?

sethmlarson:
jenkins test this please

tests/ml/test_ml_model_pytest.py
benwtrent: Yes, this should be ok. Especially if we get a ✅ build. We don't automatically translate encodings from the models anyways.

PR #364, opened 2021-07-27
Drop support for Python 3.6, pandas<1.2

sethmlarson:
Older versions are becoming harder to support, especially with pandas 1.2 dropping Python 3.6.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -53,7 +53,7 @@ TYPED_FILES = (
 def format(session):
     session.install("black", "isort")
     session.run("python", "utils/license-headers.py", "fix", *SOURCE_FILES)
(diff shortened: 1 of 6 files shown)


noxfile.py
```diff
@@ -89,20 +89,13 @@ def lint(session):
             session.error("\n" + "\n".join(sorted(set(errors))))
 
 
-@nox.session(python=["3.6", "3.7", "3.8", "3.9"])
-@nox.parametrize("pandas_version", ["1.2.0", "1.3.0"])
```
V1NAY8, line 93: Should Not remove 1.2.0 , 
builds will fail and we support 1.2.0 for > 3.6

V1NAY8 (changes requested):

Issue #368, opened 2021-07-31
Switch to Point-in-Time with search_after instead of using scroll APIs


sethmlarson:
Point in time+search_after is more memory efficient however requires a `sort` option to always be present. Luckily we can add `sort: _doc` as a good default.

V1NAY8:
I am doing this.

PR #369, opened 2021-08-03
Add `iterrows()` and `itertuples()` DataFrame API, its usage is similar to `pandas`

### LINKED ISSUE #345: Can we get batch data use `df.to_pandas()`  in the case of big data?
Hello, everybody, There are one question about `DataFrame.to_pandas()` API

Can we get batch data use `df.to_pandas()`  in the case of big data?

For example:
There are 100 million rows data in Elasticsearch
Then please look at the code below
```
df = ed.DataFrame(...)         // Now, df have 100 million rows
df.to_pandas(show_progress=True) // This will very slowly, because Elasticsearch only can return 10,000 rows per query
```
Can we use like this?
```
iterator = df.to_pandas_by_batch(batch_size=1000)
for batch in iterator:
    df_batch = batch.to_pandas()
```
Is there any other way to do the same things? Thanks！

kxbin:
Related to this issues:
**Can we get batch data use df.to_pandas() in the case of big data?**  close #345

Add to_pandas_in_batch() DataFrame API

Then, We can use below code to get batch dataframe in the case of big data
```
pd_df_iterator = ed_df.to_pandas_in_batch(batch_size=1000)
for pd_df in pd_df_iterator:
    print(pd_df)
```

If there code is something wrong, please give some suggestions. Thank you!


kxbin:
> ❌ Author of the following commits did not sign a [Contributor Agreement](https://www.elastic.co/contributor-agreement):
> [58ba4ab](https://github.com/elastic/eland/commit/58ba4ab3076656125eb9f4b48b1ab2b345bb08ea)
> 
> Please, read and sign the above mentioned agreement if you want to contribute to this project

Ok，The agreement has just been signed, but the status has not been updated yet.

V1NAY8:
So, I am thinking with this #368  We will eliminate `scroll` API.
Current logic is if batch_size is less than 10k (hard coded) we take `search` else we take `scroll`

Since, `search` API accepts only 10k size at once. else we have to use pagination to fetch next set of results. 

I am thinking we can add a new parameter to `to_pandas` called `batch_size` which can take until 10k. 
If given more than that. we throw a warning saying 10k is maximum and fetch results with max batch size

Also turn existing method into iterator by default

- So, Is a different method required for this ?
- Also we need to document how this iterator has to be used. 

Once this is done. I am thinking if `Collector` wont be required after this. Since logic will be straight forward. 

@sethmlarson What do you think ?

kxbin:
> So, I am thinking with this #368 We will eliminate `scroll` API.
> Current logic is if batch_size is less than 10k (hard coded) we take `search` else we take `scroll`
> 
> Since, `search` API accepts only 10k size at once. else we have to use pagination to fetch next set of results.
> 
> I am thinking we can add a new parameter to `to_pandas` called `batch_size` which can take until 10k.
> If given more than that. we throw a warning saying 10k is maximum and fetch results with max batch size
> 
> Also turn existing method into iterator by default
> 
> * So, Is a different method required for this ?
> * Also we need to document how this iterator has to be used.
> 
> Once this is done. I am thinking if `Collector` wont be required after this. Since logic will be straight forward.
> 
> @sethmlarson What do you think ?

@V1NAY8  Thanks for the reply

I agree with you.

We can throw a warning saying 10k is maximum and fetch results with max batch size，I will modify it now.

In addition, I also considered turn existing method into iterator by default.
but will some users want to get all the data at once?

Maybe we can determine whether to return an iterator or a dataframe based on the batch_size parameters (batch_size is None or not None).
How about this?

V1NAY8:
@kxbin Don't throw a warning now in this PR. The existing code can run for both < 10k and > 10k. I will throw if needed in my other changes.

kxbin:
@V1NAY8 Okay, I'll trouble you to throw a warning.

V1NAY8:
So, we can turn internally into iterator by default. The method `to_pandas()`  which will dump all the data at once, we can change it to iterate internally and construct a df and return. 
We can expose another method just to expose iterator
Thus compatibility will be there 😛

kxbin:
> So, we can turn internally into iterator by default. The method `to_pandas()` which will dump all the data at once, we can change it to iterate internally and construct a df and return.
> We can expose another method just to expose iterator
> Thus compatibility will be there 😛

This idea is great :smiley:

sethmlarson:
Thanks @kxbin and @V1NAY8 for your interest in this feature. I discussed this with the team and I believe we're thinking it makes sense to implement [`iterrows()`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.iterrows.html) which can be transformed by the user into a "chunking/batching".

The reason we'd like to do this instead of a DataFrame chunking API is it provides compatibility with pandas and only allows two ways to view your data (either "all" the data or every row) instead of any number of views based on `batch_size`. Does this make sense?

kxbin:
> Thanks @kxbin and @V1NAY8 for your interest in this feature. I discussed this with the team and I believe we're thinking it makes sense to implement [`iterrows()`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.iterrows.html) which can be transformed by the user into a "chunking/batching".
> 
> The reason we'd like to do this instead of a DataFrame chunking API is it provides compatibility with pandas and only allows two ways to view your data (either "all" the data or every row) instead of any number of views based on `batch_size`. Does this make sense?

@sethmlarson  Thanks for the reply

Yeah, If we can be compatible with `pandas`, it will make very sense to do so. 
Because all users are very familiar with the usage of `pandas`.

We can implement the following methods：
`ed.DataFrame.iterrows()`
`ed.DataFrame.itertuples()`

Make it similar to pandas usage:
`pandas.DataFrame.iterrows()`
`pandas.DataFrame.itertuples()`

Please give me some time, I think I can finish it.  :stuck_out_tongue:

sethmlarson:
Awesome, sounds great! :muscle: Let me know if you have questions.

kxbin:
> Awesome, sounds great! 💪 Let me know if you have questions.

Okay :smiley:

eland/dataframe.py
```diff
@@ -1446,6 +1446,39 @@ def keys(self) -> pd.Index:
         """
         return self.columns
 
+    def iterrows(self) -> Iterable[Tuple[Hashable, Series]]:
```
sethmlarson, line 1449: This function should return `Iterable[Tuple[Union[str, Tuple[str, ...]], pd.Series]]`

eland/dataframe.py
```diff
@@ -1446,6 +1446,39 @@ def keys(self) -> pd.Index:
...
+        """
+        Iterate over ed.DataFrame rows as (index, ed.Series) pairs.
+
+        Returns:
```
sethmlarson, line 1453: Instead of `Returns:` should be `Yields:`

sethmlarson:
Btw you may want to wait for https://github.com/elastic/eland/pull/370 to land and base your work off of this function that's been added. It may help you out a lot!

kxbin:
> Btw you may want to wait for #370 to land and base your work off of this function that's been added. It may help you out a lot!

Thank you for your corrections and tips.

I will wait for the landing, otherwise it is possible that _es_result() will be modified at the same time, and there will be conflicts.

sethmlarson:
@kxbin I've merged #370, now it should be easier to implement this feature using `eland.operations.search_after_with_pit()`

kxbin:
> @kxbin I've merged #370, now it should be easier to implement this feature using `eland.operations.search_after_with_pit()`

Thanks for the tips.

kxbin:
@V1NAY8 @sethmlarson  Thank you for implementing the search_yield_hits() method, this method is awesome and really useful！

I added two iterator class based on this method:

They are `PandasDataFrameRowsIterator` and `PandasDataFrameTuplesIterator`, they all inherit from `PandasDataFrameIterator`.

Inside the iterator, the `pandas.DataFrame.iterrows()` and `pandas.DataFrame.itertuples()` method is directly called.

This will ensure that we are compatible with `pandas`.

By the way, The following three methods reuse the same code.
Because I am worried that it will affect other methods that reference _es_result(), I haven't extracted it yet.
`Operations.iterrows()` 
 `Operations.itertuples()`
`Operations._es_results()` 
Do we need to extract it to create a new method?

The same code:
```
query_params, post_processing = self._resolve_tasks(query_compiler)

result_size, sort_params = Operations._query_params_to_size_and_sort(
    query_params
)

script_fields = query_params.script_fields
query = Query(query_params.query)

body = query.to_search_body()
if script_fields is not None:
    body["script_fields"] = script_fields

# Only return requested field_names and add them to body
_source = query_compiler.get_field_names(include_scripted_fields=False)
body["_source"] = _source if _source else False

if sort_params:
    body["sort"] = [sort_params]
```

eland/operations.py
```diff
@@ -1509,6 +1571,49 @@ def batch_size() -> None:
...
+        return df
+
+class PandasDataFrameRowsIterator(PandasDataFrameIterator):
+    def __init__(self, query_compiler, es_results_generator, post_processing):
+        super().__init__(query_compiler, es_results_generator, post_processing)
+
+    def __next__(self):
+        es_result = next(self._es_results_generator) # This will return one es result in each iteration
+        if es_result:
+            df = self._es_results_to_pandas([es_result]) # So, it need to be converted to list
```
V1NAY8, line 1599: So, `_es_results_to_pandas` is the important method. 
Its logic is to initialize a `FieldMappingCache` on the first document. and reuse it for all the subsequent documents.

If we pass a list with one document every time, it becomes very slow. The whole reason why cache exists goes away. 

My suggestion is to Convert  `_es_results_to_pandas` into a generator itself. which yields each document, so the state is preserved. 

So, it looks like
generator (either tuples or rows) => generator(_es_results_to_pandas) -> generator (_es_results_generator) -> ES


So, in this way only required documents are generator per yield. Even though if someone wants to break the chain, it will be efficient by not querying entire index.

eland/operations.py
```diff
@@ -1196,6 +1198,66 @@ def describe(self, query_compiler: "QueryCompiler") -> pd.DataFrame:
...
+        body["_source"] = _source if _source else False
+
+        if sort_params:
+            body["sort"] = [sort_params]
+
+        es_results_generator = search_yield_hits(
+            query_compiler=query_compiler, body=body, max_number_of_hits=result_size
+        )
+
+        return PandasDataFrameTuplesIterator(query_compiler, es_results_generator, post_processing, index, name)
```
V1NAY8, line 1259: So, the above method and below have same logic, except for `return` . Can we accomplish it in single method?

eland/operations.py
```diff
@@ -1509,6 +1571,49 @@ def batch_size() -> None:
...
+class PandasDataFrameTuplesIterator(PandasDataFrameIterator):
+    def __init__(self, query_compiler, es_results_generator, post_processing, index, name):
+        super().__init__(query_compiler, es_results_generator, post_processing)
+        self._index = index
+        self._name = name
+
+    def __next__(self):
+        es_result = next(self._es_results_generator) # This will return one es result in each iteration
+        if es_result:
+            df = self._es_results_to_pandas([es_result]) # So, it need to be converted to list
```
V1NAY8, line 1613: Same as above.

eland/operations.py
```diff
@@ -1509,6 +1571,49 @@ def batch_size() -> None:
...
+    @abstractmethod
+    def __next__(self):
+        pass
+
+    def _es_results_to_pandas(self, es_results):
+        _, df = self._query_compiler._es_results_to_pandas(es_results)
+        df = Operations._apply_df_post_processing(df, self._post_processing)
+        return df
+
+class PandasDataFrameRowsIterator(PandasDataFrameIterator):
```
V1NAY8, line 1592: Pandas returns generator right. Shall we do the same ?

eland/operations.py
sethmlarson: Great suggestion, totally agree with this! :muscle:

kxbin: This is a good idea!

I try to convert `QueryCompiler._es_results_to_pandas()` into a generator itself. And did some speed tests.

I used a data set of 50,000 rows to do the test, and the results are as follows:

Before conversion:
```
ed.iterrows(),  It took a total of `2 minutes 30 seconds` after the iteration
ed.itertuples(), It took a total of `3 minutes 53 seconds` after the iteration
ed.to_pandas(), It took `15 seconds`
```
After conversion:
```
ed.iterrows(),  It took a total of `1 minutes 52 seconds` after the iteration
ed.itertuples(), It took a total of `2 minutes 11 seconds` after the iteration
ed.to_pandas(), It took `1 minutes 54 seconds`
```

kxbin: `iterrows()` and `itertuples()` are both faster.

But `to_pandas()` has slowed down, and there should be other room for optimization. 

I would like to ask if there are more suggestions, Thanks very much.

eland/operations.py
kxbin: I think we can all return `generator`, I have completed the modification

kxbin:
I think I should have succeeded in the optimization.
At present, the same 50,000 data sets, the test speed is as follows:

```
ed.iterrows(),  It took a total of `21 seconds` after the iteration
ed.itertuples(), It took a total of `25 seconds` after the iteration
ed.to_pandas(), It took `16 seconds`
```

>
> This is a good idea!
> 
> I try to convert `QueryCompiler._es_results_to_pandas()` into a generator itself. And did some speed tests.
> 
> I used a data set of 50,000 rows to do the test, and the results are as follows:
> 
> Before conversion:
> 
> ```
> ed.iterrows(),  It took a total of `2 minutes 30 seconds` after the iteration
> ed.itertuples(), It took a total of `3 minutes 53 seconds` after the iteration
> ed.to_pandas(), It took `15 seconds`
> ```
> 
> After conversion:
> 
> ```
> ed.iterrows(),  It took a total of `1 minutes 52 seconds` after the iteration
> ed.itertuples(), It took a total of `2 minutes 11 seconds` after the iteration
> ed.to_pandas(), It took `1 minutes 54 seconds`
> ```

sethmlarson:
jenkins test this please

docs/sphinx/reference/api/eland.DataFrame.iterrows.rst
```diff
@@ -0,0 +1,6 @@
+eland.DataFrame.iterrows
+======================
```
sethmlarson, line 2: nit: Make sure the `===` is the same length as the header

docs/sphinx/reference/api/eland.DataFrame.itertuples.rst
```diff
@@ -0,0 +1,6 @@
+eland.DataFrame.itertuples
+======================
```
sethmlarson, line 2: same as above

eland/query_compiler.py
```diff
@@ -147,21 +148,19 @@ def es_dtypes(self) -> pd.Series:
...
-        show_progress: bool = False,
-    ) -> "pd.Dataframe":
+        es_results: Generator[Dict[str, Any], None, None]
+    ) -> Generator["pd.Dataframe", None, None]:
         """
         Parameters
         ----------
-        results: List[Dict[str, Any]]
-            Elasticsearch results from self.client.search
+        es_results: Generator[Dict[str, Any], None, None]
```
sethmlarson, line 156: Let's type `es_results` as `Iterable[Dict[str, Any]]`

eland/operations.py
```diff
@@ -1234,17 +1303,32 @@ def _es_results(
...
+            query_compiler=query_compiler, body=body, max_number_of_hits=result_size
         )
+        df_generator = query_compiler._es_results_to_pandas(es_results_generator)
 
-        _, df = query_compiler._es_results_to_pandas(
-            results=es_results, show_progress=show_progress
-        )
-        df = self._apply_df_post_processing(df, post_processing)
-        return df
+        # Put it in the list first, and then convert it into a pandas DataFrame
```
sethmlarson, line 1311: This is good, TIL about `pd.concat()`

eland/operations.py
```diff
@@ -1234,17 +1303,32 @@ def _es_results(
...
-            results=es_results, show_progress=show_progress
-        )
-        df = self._apply_df_post_processing(df, post_processing)
-        return df
+        # Put it in the list first, and then convert it into a pandas DataFrame 
+        # which is faster than using pandas.DataFrame.append()
+        # https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.append.html
+        df_list = []
+
+        i = 0
```
sethmlarson, line 1316: Let's use a better name for this variable, maybe `df_row_count`?

eland/query_compiler.py
```diff
@@ -277,18 +272,23 @@ def _es_results_to_pandas(
             # flatten row to map correctly to 2D DataFrame
             rows.append(self._flatten_dict(row, field_mapping_cache))
 
-            if batch_size is not None:
-                if i >= batch_size:
-                    partial_result = True
-                    break
+            if i % DEFAULT_PROGRESS_REPORTING_NUM_ROWS == 0:
```
sethmlarson, line 275: I don't think we should be yielding based on a reporting setting, maybe a batch setting instead?

sethmlarson (changes requested):
This is really great for a first contribution, thanks for working on this! I have some comments below for you, however I also have a higher-level comment of I think our `search_after_hits` implementation should be yielding "batches" of documents instead of individual documents. 

This makes more sense I think because the "post-processing" phase is a lot more efficient to be called less frequently and means we'll end up creating fewer `pd.DataFrame` in the end. We'll still be able to use the current setup we're doing now in `itertuples` and `iterrows` having `yield from pd_df.itertuples()`. I can accomplish this in a separate PR.

sethmlarson:
Looks like the `lint` and `docs` jobs are failing, make sure `nox -rs format` and `nox -rs docs` pass.

sethmlarson:
@kxbin Take a look at the change in https://github.com/elastic/eland/pull/379 and adapt your PR to do essentially:

```python
def iterrows():
    ... (setup)
    for hits in _search_yield_hits(...):
        df = _es_results_to_pandas(hits)
        yield from df.iterrows()
```

Perhaps you can even encapsulate the logic within `(setup)` but maybe that can be a separate PR. Focus on getting `iterrows()` and `itertuples()` integrated first :)

th0ger:
> I am thinking we can add a new parameter to `to_pandas` called `batch_size` which can take until 10k.
> If given more than that. we throw a warning saying 10k is maximum and fetch results with max batch size

I think this would be useful, to allow raising scroll size from eland default 1000 to elastic maximum 10K.

kxbin:
> > I am thinking we can add a new parameter to `to_pandas` called `batch_size` which can take until 10k.
> > If given more than that. we throw a warning saying 10k is maximum and fetch results with max batch size
> 
> I think this would be useful, to allow raising scroll size from eland default 1000 to elastic maximum 10K.

Yeah, Really useful.

We have now adopted a better solution, we don't use batch_size now, and change to expose iterrows and itertuples API.

Then the user can implement his own batch_size in the dataframe iteration.

th0ger:
> Then the user can implement his own batch_size in the dataframe iteration.

Would you mind providing an example of how this should be called?

kxbin:
> > Then the user can implement his own batch_size in the dataframe iteration.
> 
> Would you mind providing an example of how this should be called?

Such like this:
```
ed_flights = ed.DataFrame('localhost:9200', 'flights')
batch_size = 10000

batch_series = []

for index, row in ed_flights .iterrows()
    batch_series.append(row)

    if index % batch_size == 0:
        batch_dataframe = pd.Dataframe(batch_series)
        # Then, we can use this batch_dataframe to do something we want
        batch_series = []
    
```

PR #370, opened 2021-08-03
Switch to Point-in-Time with search_after instead of using scroll APIs

### LINKED ISSUE #368: Switch to Point-in-Time with search_after instead of using scroll APIs

V1NAY8:
Closes #368 

@sethmlarson Take a look

### DIFF
--- a/eland/operations.py
+++ b/eland/operations.py
@@ -33,13 +33,15 @@ from typing import (
 
 import numpy as np
 import pandas as pd  # type: ignore
-from elasticsearch.helpers import scan
+from elasticsearch.exceptions import NotFoundError
 
-from eland.actions import PostProcessingAction, SortFieldAction
+from eland.actions import PostProcessingAction
 from eland.common import (
     DEFAULT_CSV_BATCH_OUTPUT_SIZE,
     DEFAULT_ES_MAX_RESULT_WINDOW,
+    DEFAULT_KEEP_ALIVE,
     DEFAULT_PAGINATION_SIZE,
+    DEFAULT_SEARCH_SIZE,
     SortOrder,
     build_pd_series,
     elasticsearch_date_to_pandas_date,
@@ -1224,7 +1226,7 @@ class Operations:
     ) -> None:
         query_params, post_processing = self._resolve_tasks(query_compiler)
 
-        size, sort_params = Operations._query_params_to_size_and_sort(query_params)
+        result_size, sort_params = self._query_params_to_size_and_sort(query_params)
 
         script_fields = query_params.script_fields
         query = Query(query_params.query)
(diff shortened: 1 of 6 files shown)


sethmlarson:
jenkins test this please

sethmlarson:
jenkins test this please

eland/operations.py
```diff
@@ -1233,58 +1233,55 @@ def _es_results(
...
+            es_results.extend(results)
+            # If results is empty, we completed searching
+            if results and is_pagination_required:
+                body["search_after"] = results[-1]["sort"]
+            else:
+                break
+
+        try:
+            res = query_compiler._client.close_point_in_time(body={"id": pit_id})
+            assert res["succeeded"] is True
```
sethmlarson, line 1276: No need to assert anything here, we tried our best and the PIT will expire after the keep_alive anyways.

eland/operations.py
```diff
@@ -1233,58 +1233,55 @@ def _es_results(
...
         else:
-            partial_result, df = query_compiler._es_results_to_pandas(es_results)
-            df = self._apply_df_post_processing(df, post_processing)
-            collector.collect(df)
+            body["size"] = DEFAULT_ES_MAX_RESULT_WINDOW
+            is_pagination_required = True
+
+        if sort_params is not None:
+            body["sort"] = [sort_params]
+        elif is_pagination_required:
```
sethmlarson, line 1250: Since we're always paginating with PIT we should unconditionally add `body[sort]`, no need to do `is_pagination_required` conditional branches.

eland/operations.py
```diff
@@ -1233,58 +1233,55 @@ def _es_results(
...
+            body["sort"] = [{"_doc": "asc"}]
+
+        # To Improve performance
+        body["track_total_hits"] = False
+
+        # Initialize Point in time API
+        pit_id: str = query_compiler._client.open_point_in_time(
```
sethmlarson, line 1257: After we receive a `pit_id` we should start a `try-finally` where the `pit_id` is used so we can "guarantee" that regardless of errors within this span we attempt to release the PIT. Basically the `try-close_point_in_time` block would go under the `finally` block.

eland/operations.py
```diff
@@ -1233,58 +1233,55 @@ def _es_results(
...
+            body["size"] = DEFAULT_ES_MAX_RESULT_WINDOW
+            is_pagination_required = True
+
+        if sort_params is not None:
+            body["sort"] = [sort_params]
+        elif is_pagination_required:
+            body["sort"] = [{"_doc": "asc"}]
+
+        # To Improve performance
+        body["track_total_hits"] = False
```
sethmlarson, line 1254: This is a great idea!

eland/operations.py
```diff
@@ -1233,58 +1233,55 @@ def _es_results(
...
+        # Initialize Point in time API
+        pit_id: str = query_compiler._client.open_point_in_time(
+            index=query_compiler._index_pattern, keep_alive="1m"
+        )["id"]
+        body["pit"] = {"id": pit_id, "keep_alive": "1m"}
+
+        es_results: List[Dict[str, Any]] = []
+
+        # Check result_size to handle df.head(0) and df.tail(0)
+        while True and result_size != 0:
```
sethmlarson, line 1265: No need for `True and X`, it is equivalent to `X`

eland/operations.py
```diff
@@ -1233,58 +1233,55 @@ def _es_results(
...
+            body["sort"] = [sort_params]
+        elif is_pagination_required:
+            body["sort"] = [{"_doc": "asc"}]
+
+        # To Improve performance
+        body["track_total_hits"] = False
+
+        # Initialize Point in time API
+        pit_id: str = query_compiler._client.open_point_in_time(
+            index=query_compiler._index_pattern, keep_alive="1m"
```
sethmlarson, line 1258: Let's define `keep_alive` in one place and use everywhere, and let's set it to something larger like 5m since we're actively cleaning it up at the end of our query anyways.

sethmlarson (changes requested):
This is great, as we discussed  we should move a lot of this logic into a "scan-like' helper which yields individual documents so it can be re-used in the future when we have APIs that also yield individual documents like `iterrows`.

eland/operations.py
V1NAY8: So, I am keeping it as "3m" because. ES Documentation says following:
<html>
<body>


The keep_alive parameter tells Elasticsearch how long it should extend the time to live of the point in time.
--



</body>
</html>

So, for every search it extends the time. :)

sethmlarson: Works for me :)

sethmlarson:
jenkins test this please

eland/operations.py
```diff
@@ -1233,58 +1235,35 @@ def _es_results(
...
-                client=query_compiler._client,
-                index=query_compiler._index_pattern,
-                query=body,
+            body["size"] = DEFAULT_SEARCH_SIZE
+            is_pagination_required = True
+
+        if sort_params:
+            body["sort"] = [sort_params]
+
+        es_results = list(
```
sethmlarson, line 1253: Move the `list(...)` to only wrap the `search_after_with_pit` call.

eland/operations.py
```diff
@@ -1453,7 +1431,7 @@ def es_info(self, query_compiler: "QueryCompiler", buf: TextIO) -> None:
         buf.write(f" tasks: {self._tasks}\n")
 
         query_params, post_processing = self._resolve_tasks(query_compiler)
-        size, sort_params = Operations._query_params_to_size_and_sort(query_params)
+        size, sort_params = self._query_params_to_size_and_sort(query_params)
```
sethmlarson, line 1434: Keep this as `Operations.` since it's a static method.

eland/operations.py
```diff
@@ -1541,3 +1519,63 @@ def batch_size() -> None:
     @property
     def show_progress(self) -> bool:
         return self._show_progress
+
+
+def search_after_with_pit(
+    query_compiler: "QueryCompiler",
+    body: Dict[str, Any],
+    is_pagination_required: bool = True,
```
sethmlarson, line 1527: Remove `is_pagination_required` parameter

eland/operations.py
```diff
@@ -1541,3 +1519,63 @@ def batch_size() -> None:
...
+        An instance of query_compiler
+    body:
+        body for search API
+    is_pagination_required: bool
+        set True if size of documents to be queried is greater than 10,000 else False.
+        Defaults to True
+
+    Examples
+    --------
+    results = list(search_after_all(query_compiler, body, True))
```
sethmlarson, line 1545: Is this example complete?

eland/operations.py
```diff
@@ -1541,3 +1519,63 @@ def batch_size() -> None:
...
+
+    """
+    try:
```
sethmlarson, line 1548: Define `pit_id: Optional[str] = None` outside of the `try:`

eland/operations.py
```diff
@@ -1541,3 +1519,63 @@ def batch_size() -> None:
...
+                if is_pagination_required:
+                    body["search_after"] = results[-1]["sort"]
+                else:
+                    # This is important otherwise it fetches entire index
+                    # if pagination is not required. One search is enough
+                    break
+            else:
+                break
+    finally:
+        try:
```
sethmlarson, line 1575: Here add an `if pit_id is not None:` before the `try-except NotFoundError`

eland/operations.py
```diff
@@ -1541,3 +1519,63 @@ def batch_size() -> None:
...
+            query_compiler._client.close_point_in_time(body={"id": pit_id})
+        except NotFoundError:
+            # if pit_id is already closed, ES throws NotFoundError
+            # Hopefully It shouldn't come here
+            pass
+        return
```
sethmlarson, line 1581: Remove this `return`, it shouldn't be needed.

eland/query_compiler.py
```diff
@@ -252,7 +252,7 @@ def _es_results_to_pandas(
         rows = []
         index = []
         if isinstance(results, dict):
-            iterator = results["hits"]["hits"]
+            iterator = results
```
sethmlarson, line 255: Shouldn't this be `iterator = [results]`?

sethmlarson (changes requested):
This is epic! Thank you for this, here's some comments for you:

sethmlarson:
jenkins test this please

sethmlarson:
jenkins test this please

Issue #251, opened 2020-08-11
Start testing `reg:pseudohubererror` when XGBoost v1.2 is released


sethmlarson:
See https://github.com/dmlc/xgboost/issues/5698#issuecomment-632783978
Should be available in 1.2 which is in RC right now for Python.

PR #373, opened 2021-08-06
Optimize aggregate_field_name method in field_mappings

V1NAY8:
The method `aggregatable_field_name` has `self._mappings_capabilities` which is a dataframe.

We are querying `dataframe`  4 times per method call. Instead query it once and reuse the result.

Also this improves readability of the code.

@sethmlarson  take a look

### DIFF
--- a/eland/field_mappings.py
+++ b/eland/field_mappings.py
@@ -591,7 +591,7 @@ class FieldMappings:
         except KeyError:
             raise KeyError(
                 f"Can not get aggregatable field name for invalid display name {display_name}"
-            )
+            ) from None
 
         if mapping is not None and mapping.aggregatable_es_field_name is None:
             warnings.warn(f"Aggregations not supported for '{display_name}'")
--- a/eland/field_mappings.py
+++ b/eland/field_mappings.py
@@ -584,21 +584,19 @@ class FieldMappings:
 
         raise KeyError if the field_name doesn't exist in the mapping, or isn't aggregatable
         """
-        if display_name not in self._mappings_capabilities.index:
+        mapping: Optional[pd.Series] = None
+
+        try:
+            mapping = self._mappings_capabilities.loc[display_name]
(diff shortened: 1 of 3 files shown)


sethmlarson:
jenkins test this please

V1NAY8:
Now, this also closes #251

sethmlarson:
@V1NAY8 Could you separate these two feature contributions into different PRs? It's best to keep them separate so they don't get blocked on each other. I'd also like to have the ML change reviewed by someone else.

eland/field_mappings.py
```diff
@@ -584,21 +584,19 @@ def aggregatable_field_name(self, display_name: str) -> Optional[str]:
 
         raise KeyError if the field_name doesn't exist in the mapping, or isn't aggregatable
         """
-        if display_name not in self._mappings_capabilities.index:
+        mapping: Optional[pd.Series] = None
+
+        try:
+            mapping = self._mappings_capabilities.loc[display_name]
+        except KeyError:
             raise KeyError(
```
sethmlarson, line 592: Let's add `from None` to this `raise` so our traceback doesn't change ie: `raise KeyError("...long message") from None`

eland/query_compiler.py
```diff
@@ -799,7 +799,11 @@ def get_arithmetic_op_fields(self) -> Optional["ArithmeticOpFieldsTask"]:
 
     def display_name_to_aggregatable_name(self, display_name: str) -> Optional[str]:
```
sethmlarson, line 800: This type is no longer correct, only returns `str` not `Optional[str]`

sethmlarson (changes requested):
Thanks for this, a couple comments for you!

sethmlarson:
jenkins test this please

sethmlarson (approved):
LGTM, thanks!

PR #378, opened 2021-08-11
Remove `PandasToCSVCollector`, `PandasToDataFrameCollector` 

V1NAY8:
This PR is a small change that removes `PandasToCSVCollector`, `PandasToDataFrameCollector` 

This has a basic functionality, But at the moment I think its not required. 

If any additional functionality comes, I think we can achieve it without a class!

@sethmlarson  Take a look at it :)

### DIFF
--- a/eland/etl.py
+++ b/eland/etl.py
@@ -526,29 +526,18 @@ def csv_to_eland(  # type: ignore
 
     first_write = True
     for chunk in reader:
-        if first_write:
-            pandas_to_eland(
-                chunk,
-                es_client,
-                es_dest_index,
-                es_if_exists=es_if_exists,
-                chunksize=chunksize,
-                es_refresh=es_refresh,
-                es_dropna=es_dropna,
-                es_type_overrides=es_type_overrides,
-            )
-            first_write = False
-        else:
-            pandas_to_eland(
-                chunk,
-                es_client,
-                es_dest_index,
-                es_if_exists="append",
-                chunksize=chunksize,
-                es_refresh=es_refresh,
-                es_dropna=es_dropna,
-                es_type_overrides=es_type_overrides,
-            )
+        pandas_to_eland(
+            chunk,
+            es_client,
+            es_dest_index,
+            chunksize=chunksize,
+            es_refresh=es_refresh,
+            es_dropna=es_dropna,
+            es_type_overrides=es_type_overrides,
+            # es_if_exists should be 'append' except on the first call to pandas_to_eland()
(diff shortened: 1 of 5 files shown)


sethmlarson:
jenkins test this please

eland/etl.py
```diff
@@ -526,29 +526,17 @@ def csv_to_eland(  # type: ignore
...
-                chunk,
-                es_client,
-                es_dest_index,
-                es_if_exists="append",
-                chunksize=chunksize,
-                es_refresh=es_refresh,
-                es_dropna=es_dropna,
-                es_type_overrides=es_type_overrides,
-            )
+        pandas_to_eland(
```
sethmlarson, line 529: This is super great! :muscle: 

One nit is that we should add a comment on the `es_if_exists` line (`# es_if_exists should be 'append' except on the first call to pandas_to_eland()`?) so it stands out from all the other lines. Maybe moving it to the bottom of the function call would help as well with readability.

eland/operations.py
```diff
@@ -1198,32 +1197,23 @@ def describe(self, query_compiler: "QueryCompiler") -> pd.DataFrame:
...
         **kwargs: Union[bool, str],
     ) -> None:
-
-        collector = PandasToCSVCollector(show_progress, **kwargs)
-
-        self._es_results(query_compiler, collector)
-
-        return collector._ret
+        df = self._es_results(query_compiler, show_progress)
+        df.to_csv(**kwargs)
```
sethmlarson, line 1212: `return df.to_csv()`? We should add a type and test for this.

sethmlarson (changes requested):
This is a good change, I've got a few comments for you.

Generally this seems like an in-between change towards making `_es_results` a true iterator but that can still be handled in a separate change too.

eland/operations.py
V1NAY8: So, we actually won't return this, the path of the file is passed in kwargs, where the csv has to be ingested to. 

We already have good tests in `tests\dataframe\test_to_csv_pytest.py`
What do you think ?

sethmlarson: From reading [`pandas.DataFrame.to_csv()`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html) it looks like if `path_of_buf` is `None` then the CSV is returned as a string. We should probably do and test that as well.

sethmlarson:
jenkins test this please

sethmlarson (approved):
LGTM, thanks for this! Simplifies this logic immensely.

PR #389, opened 2021-09-16
Improve performance with efficient use of FieldMappingCache

V1NAY8:
This small change will improve the performance of `FieldMappingCache`

Existing drawbacks:
- `field_mapping_cache` is initialized on every yield of `itertuples` and `iterrows`
- Whenever eland dataframe is used to `repr()` or `print()` the `field_mapping_cache` is re initialized on every call.

Now, `field_mapping_cache` is initialized once per `query_compiler`,  that too only when ETL such as `to_pandas()` or `itertuples` or `iterrows` or `repr()` or `print()` is done.

Also, by default `scripted_fields` are considered as "object". Instead of finding source field, then if it's not source field,  an exception is raised and then its data type is assigned as "object".

This also improves the readability of code.

@sethmlarson  Take a look :)


sethmlarson:
jenkins test this please

sethmlarson:
jenkins test this please

V1NAY8:
Ok, So, the tests are good. So, a simple rebase should be fine.

sethmlarson:
jenkins test this please

PR #391, opened 2021-09-21
Add flynt and add more excluded lines to improve coverage

V1NAY8:
This PR will bump test coverage from 90% to 91% by excluding `NotImplementedError*`

Also we add `flynt` to nox format to ensure we use f-strings

@sethmlarson  take a look and hit Jenkins once.


V1NAY8:
Cool, CI is good

Issue #327, opened 2020-11-16
Improve test coverage for eland


V1NAY8:
Look at the latest main branch builds for Eland, check the `pytest-cov` results to see what lines test coverage is missing.

PR #392, opened 2021-09-22
Improve coverage for dataframe.py

V1NAY8:
For #327 

This PR will increase the coverage by reducing the misses from 309 to 295. 

@sethmlarson Take a look :)

### DIFF
--- a/tests/dataframe/test_drop_pytest.py
+++ b/tests/dataframe/test_drop_pytest.py
@@ -15,17 +15,24 @@
 #  specific language governing permissions and limitations
 #  under the License.
 
+import pytest
+
 # File called _pytest for PyCharm compatability
+from tests.common import TestData
 
 
-class TestDataFrameDrop:
+class TestDataFrameDrop(TestData):
     def test_drop(self, df):
-        df.drop(["Carrier", "DestCityName"], axis=1)
+        df.drop(labels=["Carrier", "DestCityName"], axis=1)
         df.drop(columns=["Carrier", "DestCityName"])
 
-        df.drop(["1", "2"])
-        df.drop(["1", "2"], axis=0)
+        df.drop(labels=["1", "2"])
+        df.drop(labels=["1", "2"], axis=0)
         df.drop(index=["1", "2"])
+        df.drop(labels="3", axis=0)
+        df.drop(columns="Carrier")
+        df.drop(columns=["Carrier", "Carrier_1"], errors="ignore")
+        df.drop(columns=["Carrier_1"], errors="ignore")
 
     def test_drop_all_columns(self, df):
         all_columns = list(df.columns)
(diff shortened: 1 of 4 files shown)


sethmlarson:
jenkins test this please

tests/dataframe/test_drop_pytest.py
```diff
@@ -50,3 +57,30 @@ def test_drop_all_index(self, df):
         ):
             assert dropped.shape == (0, cols)
             assert list(dropped.to_pandas().index) == []
+
+    def test_drop_raises(self):
+        ed_flights = self.ed_flights()
+
+        with pytest.raises(
```
sethmlarson, line 64: You need to have a separate `with pytest.raises(...)` for every statement you're expecting an error from.

tests/dataframe/test_drop_pytest.py
```diff
@@ -15,17 +15,24 @@
...
+
 # File called _pytest for PyCharm compatability
+from tests.common import TestData
 
 
-class TestDataFrameDrop:
+class TestDataFrameDrop(TestData):
     def test_drop(self, df):
-        df.drop(["Carrier", "DestCityName"], axis=1)
+        df.drop(labels=["Carrier", "DestCityName"], axis=1)
```
sethmlarson, line 26: Can we leave at least one without `labels=`.

tests/dataframe/test_drop_pytest.py
```diff
@@ -15,17 +15,24 @@
...
 
-        df.drop(["1", "2"])
-        df.drop(["1", "2"], axis=0)
+        df.drop(labels=["1", "2"])
+        df.drop(labels=["1", "2"], axis=0)
         df.drop(index=["1", "2"])
+        df.drop(labels="3", axis=0)
+        df.drop(columns="Carrier")
+        df.drop(columns=["Carrier", "Carrier_1"], errors="ignore")
+        df.drop(columns=["Carrier_1"], errors="ignore")
```
sethmlarson, line 36: Can we add an `assert` at the end of this test.

tests/dataframe/test_repr_pytest.py
```diff
@@ -241,6 +241,14 @@ def test_num_rows_repr_html(self):
         # Restore default
         pd.set_option("display.show_dimensions", show_dimensions)
 
+    def test_num_rows_repr_html_display_none(self):
+        display = pd.get_option("display.notebook_repr_html")
```
sethmlarson, line 245: Put this whole test in a `try:` after `display = ...` and the "restore default" step within a `finally:`

sethmlarson (changes requested):
Looks pretty good, thanks! Few comments for you:

tests/dataframe/test_drop_pytest.py
V1NAY8: Yes

tests/dataframe/test_drop_pytest.py
V1NAY8: So `conftest.py` takes care of it. It takes the df operation and applies it to both pandas and eland then asserts it. In this method. 
https://github.com/elastic/eland/blob/f241ae971aeb332c05c77fa3321b70175ea057e8/tests/conftest.py#L88

Am I missing something here ?

tests/dataframe/test_repr_pytest.py
V1NAY8: Yes :) there are some more places, I added them too

sethmlarson:
jenkins test this please

sethmlarson (approved):
LGTM, thanks for this!!

PR #393, opened 2021-09-22
Refactor `df.info()` for better readability

V1NAY8:
This PR optimizes `df.info` where `self.columns` and its `len()` is called like so many times, which always calculates and fetches it back. 

So, call it once, re use it

Also converted `_get_columns` directly to `columns` method with property.

@sethmlarson  take a look :)

### DIFF
--- a/eland/dataframe.py
+++ b/eland/dataframe.py
@@ -154,7 +154,8 @@ class DataFrame(NDFrame):
             _query_compiler=_query_compiler,
         )
 
-    def _get_columns(self) -> pd.Index:
+    @property
+    def columns(self) -> pd.Index:
         """
         The column labels of the DataFrame.
 
@@ -181,8 +182,6 @@ class DataFrame(NDFrame):
         ...   dtype='object')
         """
         return self._query_compiler.columns
-
-    columns = property(_get_columns)
 
     @property
     def empty(self) -> bool:
@@ -808,7 +807,12 @@ class DataFrame(NDFrame):
         return f"{name}: {len(self)} entries{index_summary}"
 
     def info(
-        self, verbose=None, buf=None, max_cols=None, memory_usage=None, null_counts=None
+        self,
+        verbose: Optional[bool] = None,
+        buf: Optional[StringIO] = None,
+        max_cols: Optional[int] = None,
+        memory_usage: Optional[bool] = None,
+        show_counts: Optional[bool] = None,
     ) -> None:
         """
         Print a concise summary of a DataFrame.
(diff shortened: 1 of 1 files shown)


eland/dataframe.py
```diff
@@ -844,32 +848,31 @@ def info(
 
         lines = [str(type(self)), self._index_summary()]
 
-        if len(self.columns) == 0:
+        cols: pd.Index = self.columns
```
sethmlarson, line 851: nit: Can we name these variables `columns` and `number_of_columns` for better readability?

eland/dataframe.py
```diff
@@ -808,7 +807,12 @@ def _index_summary(self):
         return f"{name}: {len(self)} entries{index_summary}"
 
     def info(
-        self, verbose=None, buf=None, max_cols=None, memory_usage=None, null_counts=None
+        self,
+        verbose: Optional[bool] = None,
+        buf: Optional[StringIO] = None,
+        max_cols: Optional[int] = None,
+        memory_usage: Optional[bool] = None,
+        show_counts: Optional[bool] = None,
```
sethmlarson, line 815: The change from `null_counts` to `show_counts` isn't explained anywhere?

eland/dataframe.py
```diff
@@ -844,32 +848,31 @@ def info(
...
+
+        exceeds_info_cols = col_count > max_cols
 
         # From pandas.DataFrame
         def _put_str(s, space) -> str:
             return f"{s}"[:space].ljust(space)
 
         def _verbose_repr() -> None:
-            lines.append(f"Data columns (total {len(self.columns)} columns):")
+            lines.append(f"Data columns (total {len(cols)} columns):")
```
sethmlarson, line 875: Can we use `number_of_columns` here? Also add `nonlocal number_of_columns` at the top of the function.

sethmlarson (changes requested):
Thanks for this, probably doesn't improve performance a ton since it's not making a call to Elasticsearch? I appreciate the `@property` fix though, if you find those elsewhere please fix them too.

Few comments for you:

sethmlarson:
jenkins test this please

V1NAY8:
So, `self.columns` will always go to the capability matrix which is pandas dataframe which contains all the information of field mappings. 

I thought why query the dataframe,  every time when we can query it once and use the result again.

eland/dataframe.py
V1NAY8: To keep inline with pandas => https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.info.html

They have deprecated `null_counts` and replaced it with `show_counts`

sethmlarson:
jenkins test this please

sethmlarson (approved):
LGTM

PR #394, opened 2021-09-27
Adds PyTorch model management

joshdevins:
This is a rough (and working) draft of porting PyTorch code into eland. This is the general shape of what I'm thinking of. It splits concerns into two - (1) generic PyTorch model management in Elasticsearch, and (2) interoperability with HuggingFace transformers and model hub.

Help and comments are welcome. There's surely a lot of formatting and general Python-ification that needs to happen still so help is welcome.

### DIFF
--- a/bin/upload_hub_model.py
+++ b/bin/upload_hub_model.py
@@ -1,0 +1,80 @@
+#!venv/bin/python
+
+#  Licensed to Elasticsearch B.V. under one or more contributor
+#  license agreements. See the NOTICE file distributed with
+#  this work for additional information regarding copyright
+#  ownership. Elasticsearch B.V. licenses this file to you under
+#  the Apache License, Version 2.0 (the "License"); you may
+#  not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+# 	http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing,
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+"""
+Copies a model from the Hugging Face model hub into an Elasticsearch cluster.
+This will create local cached copies that will be traced (necessary) and
+optionally quantized before uploading to Elasticsearch. This will also check
+the task type is supported as well as the model and tokenizer types. All
+necessary configuration is uploaded along with the model.
+"""
+
+import argparse
+import elasticsearch
+import os
+import sys
+import tempfile
+
+# project library
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
+from eland.ml.pytorch.external_models import *
+from eland.ml.pytorch.pytorch_model import *
+
+DEFAULT_URL = 'http://elastic:changeme@localhost:9200'
+MODEL_HUB_URL = 'https://huggingface.co'
+
+
+def main():
+    parser = argparse.ArgumentParser(prog='upload_hub_model.py')
+    parser.add_argument('--url', default=DEFAULT_URL,
+                        help="An Elasticsearch connection URL, e.g. http://user:secret@localhost:9200")
+    parser.add_argument('--model-id', required=True,
+                        help="The model ID in the Hugging Face model hub, "
+                             "e.g. dbmdz/bert-large-cased-finetuned-conll03-english")
+    parser.add_argument('--task-type', required=True, choices=SUPPORTED_TASK_TYPES,
+                        help=f"The task type that the model will be used for.")
+    parser.add_argument('--start', action='store_true', default=False,
+                        help="Start the model deployment after uploading. Default: False")
+    args = parser.parse_args()
+
+    es = elasticsearch.Elasticsearch(args.url, timeout=300)  # 5 minute timeout
+
+    # trace and save model, then upload it from temp file
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        print("Loading HuggingFace transformer tokenizer and model")
+        hftm = HuggingFaceTransformerModel(args.model_id, args.task_type)
+        model_path, config_path, vocab_path = hftm.save(tmp_dir)
+
+        ptm = PyTorchModel(es, args.model_id)
+        ptm.stop(ignore_not_found=True)
+        ptm.delete(ignore_not_found=True)
+        print("Uploading model")
+        ptm.upload(model_path, config_path, vocab_path)
+
+    # start the deployed model
+    if args.start:
+        print("Starting model deployment")
+        if ptm.start():
+            print(f" - started: {ptm.model_id}")
+        else:
+            print(f" - failed")
+
+
+if __name__ == '__main__':
+    main()
--- a/bin/upload_hub_model.py
+++ b/bin/upload_hub_model.py
@@ -1,5 +1,3 @@
-#!venv/bin/python
-
 #  Licensed to Elasticsearch B.V. under one or more contributor
 #  license agreements. See the NOTICE file distributed with
 #  this work for additional information regarding copyright
@@ -61,9 +59,9 @@ def main():
         hftm = HuggingFaceTransformerModel(args.model_id, args.task_type)
         model_path, config_path, vocab_path = hftm.save(tmp_dir)
 
-        ptm = PyTorchModel(es, args.model_id)
-        ptm.stop(ignore_not_found=True)
-        ptm.delete(ignore_not_found=True)
+        ptm = PyTorchModel(es, hftm.es_model_id)
+        ptm.stop()
+        ptm.delete()
         print("Uploading model")
         ptm.upload(model_path, config_path, vocab_path)
 
--- a/eland/ml/pytorch/external_models.py
+++ b/eland/ml/pytorch/external_models.py
@@ -363,7 +363,7 @@ class HuggingFaceTransformerModel:
 
     @staticmethod
     def dict_to_ordered_list(
-        dict: Dict[Any, Any], sort_by_key: bool = False, sort_by_value: bool = False
+        input_dict: Dict[Any, Any], sort_by_key: bool = False, sort_by_value: bool = False
     ) -> List[Any]:
         assert not (sort_by_key and sort_by_value)
         assert sort_by_key or sort_by_value
@@ -375,7 +375,7 @@ class HuggingFaceTransformerModel:
             sort_index = 1
             select_index = 0
 
-        items = list(dict.items())
+        items = list(input_dict.items())
         items.sort(key=lambda x: x[sort_index])
         new_list = [x[select_index] for x in items]
         return new_list
--- a/eland/ml/pytorch/external_models.py
+++ b/eland/ml/pytorch/external_models.py
@@ -27,7 +27,7 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 
 DEFAULT_OUTPUT_KEY = 'sentence_embedding'
 SUPPORTED_TASK_TYPES = {'fill_mask', 'ner', 'text_classification', 'text_embedding'}
-SUPPORTED_TASK_TYPES_STR = ', '.join(SUPPORTED_TASK_TYPES)
+SUPPORTED_TASK_TYPES_NAMES = ', '.join(list(SUPPORTED_TASK_TYPES).sorted())
 SUPPORTED_TOKENIZERS = (
     transformers.BertTokenizer,
     transformers.BertTokenizerFast,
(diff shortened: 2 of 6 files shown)


joshdevins:
PS. I'm happy to just move this work into a branch in `eland` so we can collaborate more easily on this more easily. I suspect it will take some work to get it ready.

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @staticmethod
+    def _load_json(path: str) -> Dict[str, Any]:
+        with open(path, 'r') as infile:
+            return json.load(infile)
+
+    def _upload_config(self, path: str) -> bool:
+        config = PyTorchModel._load_json(path)
+        response = self._client.ml.put_trained_model(model_id=self.model_id, body=config)
+        # TODO: Check actual result code in response
```
benwtrent, line 60: I think the client raises an error if the code isn't a success. So this should be ok.

benwtrent (approved):
This looks great to me. 

But, I am no Pythonista. Would be good to get a review from @sethmlarson

sethmlarson:
@joshdevins I like your idea of developing this on a branch, I think we can create a branch right off of `main` and then change the target branch for this PR without much fuss. Reviewing this PR now.

eland/ml/pytorch/external_models.py
```diff
@@ -0,0 +1,391 @@
...
+            # Make classes like I-PER into I_PER which fits Java enumerations
+            return [x.replace('-', '_') for x in labels]
+
+    class FillMaskModel(TraceableModel):
+        def trace(self) -> TracedModelTypes:
+            # model needs to be in evaluate mode
+            self._model.eval()
+
+            # tokenizing dummy input text
+            text = "[CLS] Who was Jim Henson ? [SEP] [MASK] Henson was a puppeteer [SEP]"
```
sethmlarson, line 204: This looks like real data, are these supposed to be parameters?

eland/ml/pytorch/external_models.py
```diff
@@ -0,0 +1,391 @@
...
+        if not isinstance(self._tokenizer, SUPPORTED_TOKENIZERS):
+            raise TypeError(f"Tokenizer type {self._tokenizer} not supported, must be one of: {SUPPORTED_TOKENIZERS_STR}")
+
+        self._traceable_model = self._create_traceable_model()
+        self._traced_model = self._traceable_model.trace()
+        self._vocab = self._load_vocab()
+        self._config = self._create_config()
+
+    @staticmethod
+    def _dict_to_ordered_list(dict: Dict[Any, Any], sort_by_key: bool = False, sort_by_value: bool = False) -> List[Any]:
```
sethmlarson, line 300: Seems like a function instead of a staticmethod? Also since there are two distinct functionalities should probably be split into two?

```python
# sort_by_key
sorted_values = [v for _, v in sorted(dict.items(), key=lambda kv: kv[0])]

# sort_by_value
sorted_keys = [k for k, _ in sorted(dict.items(), key=lambda kv: kv[1])]
```

eland/ml/pytorch/external_models.py
```diff
@@ -0,0 +1,391 @@
...
+        return new_list
+
+    def _load_vocab(self):
+        vocabulary = HuggingFaceTransformerModel._dict_to_ordered_list(self._tokenizer.get_vocab(), sort_by_value=True)
+        return {
+            'vocabulary': vocabulary,
+        }
+
+    def _create_config(self):
+        do_lower_case = hasattr(self._tokenizer, 'do_lower_case') and self._tokenizer.do_lower_case
```
sethmlarson, line 323: This is more readable as `getattr(self._tokenizer, "do_lower_case", False)`

eland/ml/pytorch/external_models.py
```diff
@@ -0,0 +1,391 @@
...
+            'model_type': 'pytorch',
+            'inference_config': inference_config,
+            'input': {
+                'field_names': ['text_field'],
+            }
+        }
+
+    def _create_traceable_model(self) -> TraceableModel:
+
+        if self._task_type == 'fill_mask':
```
sethmlarson, line 354: nit: Some spacing between each of the `if-elif` blocks here would do a lot for readability.

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+
+    def __init__(
+        self,
+        es_client: Union[str, List[str], Tuple[str, ...], "Elasticsearch"],
+        model_id: str,
+    ):
+        self._client = ensure_es_client(es_client)
+        # Elasticsearch model IDs need to be a specific format: no special chars, all lowercase, max 64 chars
+        self.model_id = model_id.replace('/', '__').lower()[:64]
```
sethmlarson, line 50: Instead of silently trimming/cleaning up we should raise a `ValueError` with these requirements.

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,122 @@
...
+
+        def model_file_chunk_generator():
+            with open(model_path, 'rb') as f:
+                while True:
+                    data = f.read(chunk_size)
+                    if not data:
+                        break
+                    yield base64.b64encode(data).decode()
+
+        for i, data in tqdm(enumerate(model_file_chunk_generator(), start=0), total=total_parts):
```
sethmlarson, line 80: TQDM isn't always desirable, only when we're running from CLI. I still haven't "solved" this problem for Eland in general either so not sure we should include here.

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,122 @@
...
+                'total_definition_length': file_stats.st_size,
+                'total_parts': total_parts,
+                'definition': data,
+            }
+            self._client.transport.perform_request(
+                method='PUT', url=f'/_ml/trained_models/{self.model_id}/definition/{i}', body=body)
+
+        return True
+
+    def upload(self, model_path: str, config_path: str, vocab_path: str, chunk_size: int = DEFAULT_CHUNK_SIZE) -> bool:
```
sethmlarson, line 91: Is there any cleanup we need to do if one part of the upload process passes and the next fails?

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,122 @@
...
+        else:
+            ignorables = ()
+
+        return self._client.transport.perform_request(
+            method='POST',
+            url=f'/_ml/trained_models/{self.model_id}/deployment/_stop',
+            params={'ignore': ignorables},
+        )
+
+    def delete(self, ignore_not_found=False) -> Dict[str, Any]:
```
sethmlarson, line 116: For `eland.ml.MLModel` we unconditionally ignore 404s, probably can do that here too?

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,122 @@
...
+            params={'timeout': '60s', 'wait_for': 'started'}
+        )
+
+    def stop(self, ignore_not_found=False) -> bool:
+        if ignore_not_found:
+            ignorables = 404
+        else:
+            ignorables = ()
+
+        return self._client.transport.perform_request(
```
sethmlarson, line 110: Is there any way to distinguish between the "model deployment is already stopped" and "model isn't found" case? If so might be good to use that here and ignore the "already stopped" case unconditionally and always reraise the 404 model not found case?

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,122 @@
...
+
+        return True
+
+    def upload(self, model_path: str, config_path: str, vocab_path: str, chunk_size: int = DEFAULT_CHUNK_SIZE) -> bool:
+        # TODO: Implement some pre-flight checks on config, vocab, and model
+        return self._upload_config(config_path) and \
+               self._upload_vocab(vocab_path) and \
+               self._upload_model(model_path, chunk_size)
+
+    def start(self) -> bool:
```
sethmlarson, line 97: Timeout should probably be configurable.

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,122 @@
...
+
+if TYPE_CHECKING:
+    from elasticsearch import Elasticsearch
+
+DEFAULT_CHUNK_SIZE = 4 * 1024 * 1024  # 4MB
+QUEUE_SIZE = 4
+THREAD_COUNT = 4
+
+
+class PyTorchModel:
```
sethmlarson, line 34: Are we siloing this to be only PyTorch?

sethmlarson (changes requested):
This is very exciting! I left a whole bunch of review comments. a high level review comment is you can run `nox -rs format` (or `black eland/` will do a lot too) to automatically format code to be compliant with the linter. You'll also get type-checking as well.

joshdevins:
Ok, let me go through the comments here and I will get this into a cleaner state. I'll then push the branch from my fork to this repo so we can collaborate on a branch here until we're ready to merge into master.

bin/upload_hub_model.py
```diff
@@ -0,0 +1,80 @@
+#!venv/bin/python
```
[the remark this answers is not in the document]
benwtrent: its an example of how to download a model from hugging face and then deploy it. Its a useful script that the user can run. 

I don't think there is a plan to have it installed on the users system.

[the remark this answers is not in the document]
joshdevins: Sorry this was a copy-paste from the old script so it's incorrect.

I discussed this a couple months ago with @sethmlarson, we should be providing scripts for the most common use-cases so people don't need to write a notebook or their own script. The original intention is indeed that this is installed (on the system/in a venv) with `pip install eland[pytorch]`. Happy to restructure as needed. Also open to any other ideas for how this script should run though.

eland/ml/pytorch/external_models.py
```diff
@@ -0,0 +1,391 @@
...
+
+        def classification_labels(self) -> Optional[List[str]]:
+            return None
+
+        @abstractmethod
+        def trace(self) -> TracedModelTypes:
+            ...
+
+    class TraceableClassificationModel(TraceableModel, ABC):
+        def classification_labels(self) -> Optional[List[str]]:
```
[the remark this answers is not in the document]
joshdevins: That's to conform with the base class which returns `None` by default

eland/ml/pytorch/external_models.py
joshdevins: Nope 😄  This does need to be improved but yes, you have to have some hard-coded examples that are used.

eland/ml/pytorch/external_models.py
joshdevins: I'm gonna do this change last, because I'm not sure it's possible. Let's come back to this.

eland/ml/pytorch/pytorch_model.py
joshdevins: Ok, this is an "automagic" conversion that we need to do, but you're right, it belongs somewhere else (in the place where the ID is automatically created from another model). 💯

joshdevins: Validation is happening already at upload time so I'll leave it there for now

sethmlarson:
@joshdevins The branch has been created and updated the PR base.

eland/ml/pytorch/pytorch_model.py
joshdevins: I know, I just don't know how to provide good feedback here to scripts since this upload can take a while. I've seen `autonotebook` used as a good way to control output.

```
from tqdm.autonotebook import tqdm
tqdm.pandas()
```

joshdevins: Or `tqdm.auto` 🤔

eland/ml/pytorch/pytorch_model.py
joshdevins: @dimitris-athanasiou ^^ ?

eland/ml/pytorch/pytorch_model.py
joshdevins: @benwtrent ^^ ?

eland/ml/pytorch/pytorch_model.py
joshdevins: Yes, for now it's all we support.

eland/ml/pytorch/external_models.py
```diff
@@ -0,0 +1,391 @@
...
+        return self._model(**{
+            'input_ids': input_ids,
+            'attention_mask': attention_mask,
+            'token_type_ids': token_type_ids,
+        })
+
+
+class HuggingFaceTransformerModel:
+
+    class TraceableModel(ABC):
```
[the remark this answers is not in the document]
joshdevins: I've made them normal classes but it might be work it's own module. I'm not sure I love as it is now.

eland/ml/pytorch/external_models.py
joshdevins: This is called from outside the class as well so I'm gonna leave it for now. Happy to review once this is on a branch, if there is a better way.

eland/ml/pytorch/pytorch_model.py
benwtrent: Looking at the code, `_stop` model deployment doesn't throw when the model is missing or when the deployment is already stopped.

PR #395, opened 2021-09-29
Adds PyTorch model management

joshdevins:
This adds PyTorch model management to eland. This is a clean branch with changes made from our last discussions. The main thing to continue working on in this PR is how we expose the script `upload_hub_model.py` since it should be "installed" and easily run with a simple command. Happy to discuss other items inline as well.

### DIFF
--- a/eland/ml/pytorch/pytorch_model.py
+++ b/eland/ml/pytorch/pytorch_model.py
@@ -1,0 +1,140 @@
+#  Licensed to Elasticsearch B.V. under one or more contributor
+#  license agreements. See the NOTICE file distributed with
+#  this work for additional information regarding copyright
+#  ownership. Elasticsearch B.V. licenses this file to you under
+#  the Apache License, Version 2.0 (the "License"); you may
+#  not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+# 	http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing,
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+import base64
+import json
+import math
+import os
+from typing import TYPE_CHECKING, Any, Dict, List, Tuple, Union
+
+import elasticsearch
+from tqdm.auto import tqdm
+
+from eland.common import ensure_es_client
+
+if TYPE_CHECKING:
+    from elasticsearch import Elasticsearch
+
+DEFAULT_CHUNK_SIZE = 4 * 1024 * 1024  # 4MB
+QUEUE_SIZE = 4
+THREAD_COUNT = 4
+DEFAULT_TIMEOUT = "60s"
+
+
+class PyTorchModel:
+    """
+    A PyTorch model managed by Elasticsearch.
+
+    These models must be trained outside of Elasticsearch, conform to the
+    support tokenization and inference interfaces, and exported as their
+    TorchScript representations.
+    """
+
+    def __init__(
+        self,
+        es_client: Union[str, List[str], Tuple[str, ...], "Elasticsearch"],
+        model_id: str,
+    ):
+        self._client = ensure_es_client(es_client)
+        self.model_id = model_id
+
+    @staticmethod
+    def _load_json(path: str) -> Dict[str, Any]:
+        with open(path, "r") as infile:
+            return json.load(infile)
+
+    def _upload_config(self, path: str) -> bool:
+        config = PyTorchModel._load_json(path)
+        response = self._client.ml.put_trained_model(
+            model_id=self.model_id, body=config
+        )
+        return response is not None
+
+    def _upload_vocab(self, path: str) -> bool:
+        vocab = PyTorchModel._load_json(path)
+        return self._client.transport.perform_request(
+            method="PUT",
+            url=f"/_ml/trained_models/{self.model_id}/vocabulary",
+            body=vocab,
+        )
+
+    def _upload_model(
+        self, model_path: str, chunk_size: int = DEFAULT_CHUNK_SIZE
+    ) -> bool:
+        file_stats = os.stat(model_path)
+        total_parts = math.ceil(file_stats.st_size / chunk_size)
+
+        def model_file_chunk_generator():
+            with open(model_path, "rb") as f:
+                while True:
+                    data = f.read(chunk_size)
+                    if not data:
+                        break
+                    yield base64.b64encode(data).decode()
+
+        for i, data in tqdm(
+            enumerate(model_file_chunk_generator(), start=0), total=total_parts
+        ):
+            body = {
+                "total_definition_length": file_stats.st_size,
+                "total_parts": total_parts,
+                "definition": data,
+            }
+            self._client.transport.perform_request(
+                method="PUT",
+                url=f"/_ml/trained_models/{self.model_id}/definition/{i}",
+                body=body,
+            )
+
+        return True
+
+    def upload(
+        self,
+        model_path: str,
+        config_path: str,
+        vocab_path: str,
+        chunk_size: int = DEFAULT_CHUNK_SIZE,
+    ) -> bool:
+        # TODO: Implement some pre-flight checks on config, vocab, and model
+        return (
+            self._upload_config(config_path)
+            and self._upload_vocab(vocab_path)
+            and self._upload_model(model_path, chunk_size)
+        )
+
+    def start(self, timeout: str = DEFAULT_TIMEOUT) -> bool:
+        return self._client.transport.perform_request(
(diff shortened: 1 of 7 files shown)


eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+        # TODO: Implement some pre-flight checks on config, vocab, and model
+        return (
+            self._upload_config(config_path)
+            and self._upload_vocab(vocab_path)
+            and self._upload_model(model_path, chunk_size)
+        )
+
+    def start(self, timeout: str = DEFAULT_TIMEOUT) -> bool:
+        return self._client.transport.perform_request(
+            method="POST",
```
sethmlarson, line 121: Let's move the `method` and `url` options to be positional:

```python
self._client.transport.perform_request("POST", f"/_ml/...", params={...})
```

Also the return type here is `bool` but this function either raises an exception or returns a JSON body. Perhaps we should switch to `-> None` and drop the `return`?

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+            url=f"/_ml/trained_models/{self.model_id}/deployment/_start",
+            params={"timeout": timeout, "wait_for": "started"},
+        )
+
+    def stop(self) -> bool:
+        try:
+            return self._client.transport.perform_request(
+                method="POST",
+                url=f"/_ml/trained_models/{self.model_id}/deployment/_stop",
```
sethmlarson, line 130: Same as above. Also let's remove the `ignore: 404` as we're catching the exception.

Can we remove the return and return type on this method as well?

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+                params={"ignore": 404},
+            )
+        except elasticsearch.NotFoundError:
+            pass
+
+    def delete(self) -> Dict[str, Any]:
```
sethmlarson, line 136: Same comments as above

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+        try:
+            return self._client.ml.delete_trained_model(self.model_id, ignore=(404,))
+        except elasticsearch.NotFoundError:
+            pass
+
+    @staticmethod
+    def list(
```
sethmlarson, line 143: Since this isn't doing a ton beyond what the client is already doing, thoughts on removing this method?

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+            }
+            self._client.transport.perform_request(
+                method="PUT",
+                url=f"/_ml/trained_models/{self.model_id}/definition/{i}",
+                body=body,
+            )
+
+        return True
+
+    def upload(
```
sethmlarson, line 105: Let's switch to a `@classmethod` called `import_model` to match with the existing `eland.ml.MLModel` class?

https://github.com/elastic/eland/blob/main/eland/ml/ml_model.py#L239

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+        self,
+        model_path: str,
```
sethmlarson, line 107: I'm not sure if it makes sense to do this but a lot of times when users are presented with an API that only takes paths to files they immediately as for "can I do it with data in memory"? Does that even make sense here, if so I have an addition review comment to ask if we can support "path or loaded JSON"

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+            body=vocab,
+        )
+
+    def _upload_model(
+        self, model_path: str, chunk_size: int = DEFAULT_CHUNK_SIZE
+    ) -> bool:
+        file_stats = os.stat(model_path)
+        total_parts = math.ceil(file_stats.st_size / chunk_size)
+
+        def model_file_chunk_generator():
```
sethmlarson, line 81: Let's add type hints to inner functions too. ` -> Iterable[str]`

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+            with open(model_path, "rb") as f:
+                while True:
+                    data = f.read(chunk_size)
+                    if not data:
+                        break
+                    yield base64.b64encode(data).decode()
+
+        for i, data in tqdm(
+            enumerate(model_file_chunk_generator(), start=0), total=total_parts
```
sethmlarson, line 90: nit: start=0 is the default so not needed

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+
+        def model_file_chunk_generator():
+            with open(model_path, "rb") as f:
+                while True:
+                    data = f.read(chunk_size)
+                    if not data:
+                        break
+                    yield base64.b64encode(data).decode()
+
+        for i, data in tqdm(
```
sethmlarson, line 89: Should we do `tqdm.auto...` here?

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+        return self._client.transport.perform_request(
+            method="PUT",
+            url=f"/_ml/trained_models/{self.model_id}/vocabulary",
+            body=vocab,
+        )
+
+    def _upload_model(
+        self, model_path: str, chunk_size: int = DEFAULT_CHUNK_SIZE
+    ) -> bool:
+        file_stats = os.stat(model_path)
```
sethmlarson, line 78: nit: Since `file_stats.st_size` appears to be the only thing used perhaps we can save that value instead of the whole stats object?

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+    def __init__(
+        self,
+        es_client: Union[str, List[str], Tuple[str, ...], "Elasticsearch"],
+        model_id: str,
+    ):
+        self._client = ensure_es_client(es_client)
+        self.model_id = model_id
+
+    @staticmethod
+    def _load_json(path: str) -> Dict[str, Any]:
```
sethmlarson, line 56: Can we inline `with open(path) as f: config=json.load(f)`?

sethmlarson (changes requested):
Collection of comments regarding only the `PytorchModel` implementation

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
+#  Licensed to Elasticsearch B.V. under one or more contributor
```
sethmlarson, line 1: Also can we make this module private like `_pytorch_model.py` so users know not to import things from here?

eland/ml/pytorch/pytorch_model.py
```diff
@@ -0,0 +1,154 @@
...
+import elasticsearch
+from tqdm.auto import tqdm
+
+from eland.common import ensure_es_client
+
+if TYPE_CHECKING:
+    from elasticsearch import Elasticsearch
+
+DEFAULT_CHUNK_SIZE = 4 * 1024 * 1024  # 4MB
+QUEUE_SIZE = 4
```
sethmlarson, line 33: Some of these variables aren't used?

eland/ml/pytorch/pytorch_model.py
joshdevins: I wanted to have start, stop, delete all in the same place since right now the client doesn't have start and stop on trained model. So having them all here is just consistent until they are all available in the client.

joshdevins: I can remove `list` since I only really use it in a "cleanup" script for QA.

eland/ml/pytorch/pytorch_model.py
joshdevins: Yes, definitely. I might need some help making this look right. In the case of model upload, I need a `BinaryIO`. For the configs, it can be easily a path to load or a `Dict` already loaded.

eland/ml/pytorch/pytorch_model.py
joshdevins: I tried this and it doesn't quite feel right. I prefer creating a `PyTorchModel` first so a user can choose to check for existing models, stop it, etc. This obviates the need for `es_exists_model` parameter like in `MLModel`.

joshdevins:
@sethmlarson Have a peek again, I've addressed most of the comments (and our Slack discussion to cleanup the scripts). There are a one or two unresolved issues that I need to chat about/get help with still.

sethmlarson (changes requested):
Few small comments then I think we can merge this and iterate on the CLI location+config in another PR.

sethmlarson:
Let's get it merged then, I can review PRs but I won't have any time to do the development myself unfortunately as I'm slammed with 8.0 work.

Issue #243, opened 2020-07-17
Add `num_samples` to transformed models (if applicable)


benwtrent:
Elastic ML utilizes Shapley Values to determine feature importance. 

For boosted regression trees, a simple form of this is to know the number of samples from the training data that "landed" in each leaf. 

Certain model formats keep track of this and it can be passed to the transformed model so that feature importance can be calculated at runtime for the model.

V1NAY8:
@benwtrent Can you give me a little guidance on how to approach/ integrate this feature? Also, an example on how it would look like.

benwtrent:
Heya @V1NAY8, so, the leaf structure in Elasticsearch machine learning has a field called:
```
number_samples
```
Having this accurately set allows for inference to supply feature importance via Shapley.

Reading: https://lightgbm.readthedocs.io/en/latest/R/reference/lgb.model.dt.tree.html

I think the correct value for light gbm is `internal_count`. Basically, how many observed examples were seen by the node. That is what should be put in the `number_samples` field for the leaf structure.

PR #398, opened 2021-09-30
Add `number_samples` to `XGBoost` ML Models

V1NAY8:
Work on #243 

TreeNode of `xgboost` has the following information:
```
Tree               1
Node               0
ID               1-0
Feature           f1
Split      -0.030468
Yes              1-1
No               1-2
Missing          1-1
Gain       44.768673
Cover      23.395073
Name: 7, dtype: object
```
So, based on 
- `importance_type` parameter of https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBRegressor
- based on https://datascience.stackexchange.com/a/12594

I am thinking `Cover` gives us the number of training samples that is being covered by the node or leaf. I might be wrong though!

@benwtrent Please give a review :)

### DIFF
--- a/eland/ml/transformers/xgboost.py
+++ b/eland/ml/transformers/xgboost.py
@@ -89,7 +89,11 @@ class XGBoostForestTransformer(ModelTransformer):
             )
 
     def build_leaf_node(self, row: pd.Series, curr_tree: int) -> TreeNode:
-        return TreeNode(node_idx=row["Node"], leaf_value=[float(row["Gain"])])
+        return TreeNode(
+            node_idx=row["Node"],
+            leaf_value=[float(row["Gain"])],
+            number_samples=int(row["Cover"]),
+        )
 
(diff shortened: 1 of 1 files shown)


eland/ml/transformers/xgboost.py
```diff
@@ -103,6 +107,7 @@ def build_tree_node(self, row: pd.Series, curr_tree: int) -> TreeNode:
                 right_child=self.extract_node_id(row["No"], curr_tree),
                 threshold=float(row["Split"]),
                 split_feature=self.get_feature_id(row["Feature"]),
+                number_samples=int(row["Cover"]),
```
benwtrent, line 110: I don't think this is gonna work. 

`cover` is a float. I don't know how they calculate `cover` related to the total number of samples that hit the tree. The docs indicate its the average of the total number of samples effected by the split. 

I wonder if `cover` * `number of features` ~= The total number of docs? Can you investigate and confirm?

PR #413, opened 2021-11-25
Adds pytorch files to noxfile

joshdevins:
We added the `pytorch` module which is type checked but was not in the noxfile as such.


noxfile.py
```diff
@@ -44,6 +44,9 @@
     "eland/ml/_optional.py",
     "eland/ml/_model_serializer.py",
     "eland/ml/ml_model.py",
+    "eland/ml/pytorch/__ini__.py",
```
sethmlarson, line 47: Typo in `__init__`

sethmlarson (approved):
LGTM, one thing to fix though:

noxfile.py
joshdevins: I just type so fast and the `t` fell off.

joshdevins:
@sethmlarson Can you have a quick pass again? I tried to go through and also fix the type warning.

eland/ml/pytorch/transformers.py
```diff
@@ -390,7 +387,7 @@ def _create_config(self):
                     "max_sequence_length"
                 ] = max_sequence_length
 
-        if self._traceable_model.classification_labels():
+        if isinstance(self._traceable_model, _TraceableClassificationModel):
```
sethmlarson, line 390: Are we losing a comparison here? Previously we also called `.classification_labels()`, do we need both or is the `isinstance` enough?

sethmlarson (approved):
LGTM, had one quick question but barring that we can merge.

eland/ml/pytorch/transformers.py
joshdevins: I was running in to mypy errors but I've fixed that so I've reverted to the previous approach.

PR #417, opened 2021-12-06
Upgrade PyTorch dependencies to latest

joshdevins:
In preparation for an 8.0 release, this updates PyTorch NLP dependencies
to more recent and latest minor versions. Amongst other things, this
introduces a fix from transformers that is helpful for text embedding
tasks with certain DPR models.

See: https://github.com/huggingface/transformers/issues/13670

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ extras = {
     "pytorch": [
         "sentence-transformers>=2.1.0,<3",
         "torch>=1.9.0,<2",
-        "transformers[torch]>=4.12.0<5",
+        "transformers[torch]>=4.12.0,<5",
     ],
 }
 extras["all"] = list({dep for deps in extras.values() for dep in deps})
--- a/setup.py
+++ b/setup.py
@@ -59,10 +59,9 @@ extras = {
     "scikit-learn": ["scikit-learn>=0.22.1,<1"],
     "lightgbm": ["lightgbm>=2,<4"],
     "pytorch": [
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -59,10 +59,9 @@
     "scikit-learn": ["scikit-learn>=0.22.1,<1"],
     "lightgbm": ["lightgbm>=2,<4"],
     "pytorch": [
-        "huggingface-hub>=0.0.17,<1",
-        "sentence-transformers>=2.0.0,<3",
+        "sentence-transformers>=2.1.0,<3",
         "torch>=1.9.0,<2",
-        "transformers[torch]>=4.11.0<5",
+        "transformers[torch]>=4.12.0<5",
```
sethmlarson, line 64: We need a comma between `4.12.0` and `<5`

sethmlarson (changes requested):
Looks good, one issue I saw just now in the `transformers[torch]` dependency:

setup.py
joshdevins: Funny, this passes linter locally 🤔

sethmlarson (approved):

joshdevins:
@sethmlarson I think the CI failures are unrelated to this change.

joshdevins:
@sethmlarson Will this make it to the 8.0 release or does it need a backport to the right branch(es)?

sethmlarson:
@joshdevins No backport needed!

PR #423, opened 2021-12-29
Add option to disable schema enforcement for `pandas_to_eland`.

### LINKED ISSUE #412: Optionally disable verify_mapping_compatibility in pandas_to_eland append
When converting pandas DataFrames and appending to ES index, the current implementation is strict and enforces compatibility between the pandas DataFrame and the ES mapping.

A use case is multiple pandas DataFrames are appended to an ES index. For example,

```
df1 = pd.DataFrame(data={'key': ['test'], 'value1': [10] })
df2 = pd.DataFrame(data={'key': ['test'], 'value2': [20] })

ed.pandas_to_eland(df1, es_client, es_index, es_dropna=True, 
                               es_if_exists='append', use_pandas_index_for_es_ids=False)

# currently throws ValueError
ed.pandas_to_eland(df2, es_client, es_index, es_dropna=True, 
                               es_if_exists='append', use_pandas_index_for_es_ids=False)
```

It can be useful to disable this strict behaviour so new dynamic mappings can be created in the ES index. This would result in:

```
ed.DataFrame(es_client, es_index)

    key  value1  value2
0  test    10.0     NaN
1  test     NaN    20.0
```

Therefore, an option is to add an argument such as `verify_mapping_compatibility: False`

Ashton-Sidhu:
Resolves #412 

Takes the suggestion by @stevedodson in #412 and adds a function argument to `pandas_to_eland` to append data where the schema in the Dataframe differs from the Elasticsearch index schema.

Added tests & documentation as well.


eland/etl.py
```diff
@@ -77,6 +78,9 @@ def pandas_to_eland(
         * False: Include missing values - may cause bulk to fail
     es_type_overrides: dict, default None
         Dict of field_name: es_data_type that overrides default es data types
+    es_enforce_index_schema: bool, default 'True'
```
sethmlarson, line 81: Let's use `es_verify_mapping_compatibility` for a name here.

Ashton-Sidhu: Sounds good, will make those changes.

Tried `es_enforce_index_schema` to see if that was preferred :p

Ashton-Sidhu: Done!

stevedodson:
jenkins test this please

PR #425, opened 2022-01-03
Add coverage for datetime columns with timezones

Ashton-Sidhu:
This PR enables pandas columns that have been cast with a timezone to be automatically be converted to date.

Without this change, the default mapping would be used and an error would be thrown.


sethmlarson:
@Ashton-Sidhu Thanks! Can you add a test case that covers this new functionality?

Ashton-Sidhu:
> @Ashton-Sidhu Thanks! Can you add a test case that covers this new functionality?
> 
> Also looks like you should run `nox -rs format` and submit the commit again for formatting.

Sorry about that, will get that added today!

sethmlarson (approved):
LGTM, thanks!

PR #432, opened 2022-02-04
Add option to disable certificate verification in eland_import_hub_model

ugosan:
With Elasticsearch 8.0.0 enabling TLS by default, it makes it hard to test eland locally since the elasticsearch-py library cannot verify the self-signed certificates. 
This PR adds the `--insecure` option to disable TLS verification

### DIFF
--- a/bin/eland_import_hub_model
+++ b/bin/eland_import_hub_model
@@ -41,7 +41,13 @@ def main():
     parser.add_argument(
         "--url",
         required=True,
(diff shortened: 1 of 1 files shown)


sethmlarson:
@ugosan Instead of adding an option to disable certificate verification should we instead add an option for specifying certificates?

ugosan:
> @ugosan Instead of adding an option to disable certificate verification should we instead add an option for specifying certificates?

We could add an option for specifying certificates *on top* of one for passing the certificate authority, IMO.

Issue #422, opened 2021-12-21
Add options to eland_import_hub_model for specifying CA certificates


ailanmar:
When uploading a Huggingface model to an ES cluster deployed through the [ML-QA framework](https://docs.elastic.dev/ml-team/docs/qa/framework) with SSL enabled (which is default) using `eland_import_hub_model ` i.e.

```
eland_import_hub_model \
--url $ELASTICSEARCH_URL \
--hub-model-id sentence-transformers/all-MiniLM-L6-v2 \
--task-type text_embedding \
--start
```

throws the following error:

```
elasticsearch.exceptions.SSLError: ConnectionError([SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1125)) caused by: SSLError([SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1125))
```

A common way to overcome this problem is to deploy the cluster disabling SSL:

```
LOCAL_CLUSTER_NAME=nlp-demo
ES_SSL_ENABLED=false KIBANA_SSL_ENABLED=false
./gradlew deployCluster startCluster
```

Here I propose an alternative way to overcome this problem in a generic way which will make `eland_import_hub_model` work in all circumstances :
Modify Line 86: 
``` python
es = elasticsearch.Elasticsearch(args.url, timeout=300)  # 5 minute timeout
```
by
``` python
try:
    es = elasticsearch.Elasticsearch(args.url, timeout=300) # 5 minute timeout
    es.info()
except elasticsearch.SSLError:
    from elasticsearch.connection import create_ssl_context
    import ssl
    ssl_context = create_ssl_context()
    ssl_context.check_hostname = False
    ssl_context.verify_mode = ssl.CERT_NONE
    es = elasticsearch.Elasticsearch(args.url, verify_certs = False, ssl_context = ssl_context, timeout=300) # 5 minute timeout
```

In this way it would be possible to establish connection to the ES host whether SSL is enabled or not and upload the hub model. I have tested this modification and works fine. If it looks good to you I will make a PR with the changes.

sethmlarson:
Instead of disabling certificate verification we should add options to specify certificates like `--ca-certs`.

RiccardoTOTI:
[...]

PR #441, opened 2022-03-01
Add --insecure and --ca-cert options to the model hub upload script

davidkyle:
When using the `eland_import_hub_model` script with https and a self signed certificate it fails with the error

```
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1131)
```

The `verify_certs` should be set to false when creating the Elasticsearch client to work-around the problem.
Using a self-signed certificate might not be unusual for on prem users.

The change also removes a deprecation warning by switching from `timeout` to `request_timeout`

### DIFF
--- a/bin/eland_import_hub_model
+++ b/bin/eland_import_hub_model
@@ -29,10 +29,11 @@ import argparse
 import tempfile
 
 import elasticsearch
+from elastic_transport.client_utils import DEFAULT
 
 from eland.ml.pytorch import PyTorchModel
-from eland.ml.pytorch.transformers import SUPPORTED_TASK_TYPES, TransformerModel
-from elastic_transport.client_utils import DEFAULT
+from eland.ml.pytorch.transformers import (SUPPORTED_TASK_TYPES,
+                                           TransformerModel)
 
 MODEL_HUB_URL = "https://huggingface.co"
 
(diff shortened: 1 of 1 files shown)


sethmlarson:
Similar to https://github.com/elastic/eland/pull/432 I'd like to first add a way to specify certificates properly instead of only allowing a method to disable certificate verification. I've captured that feature request here: https://github.com/elastic/eland/issues/422

sethmlarson:
@davidkyle You can run `nox -s format` to fix all the formatting issues.

davidkyle:
I opened #442 to address the mypy failures.

davidkyle:
@sethmlarson I had no idea the almost exact change had already been proposed in a different PR. I agree it is better to expose the `--ca-cert` option, `--insecure` was considered MVP.

A number of people have encountered this problem, not just users of the ML QA framework so I believe there is a real need to add these options. 

The `eland_import_hub_model` script was designed to reduce friction for the getting started with NLP models experience. To help those users (who may not know Python) I'd like to expose whatever options we feel are necessary.

sethmlarson (approved):

PR #442, opened 2022-03-08
Fix mypy type checker errors

davidkyle:
Corrects 2 places where the type was either wrongly inferred or wrongly defined.

There errors popped up in #441 but since they are not strictly related to that change I thought it worth a new PR


sethmlarson (approved):
LGTM, thanks!!

PR #450, opened 2022-03-31
Improve CSV Output performance

### LINKED ISSUE #449: understanding to_csv and memory usage
Hi,
I need to store an elasticsearch index, which is way larger than the memory available, to a csv file. I thought eland would do this efficiently for me under the hood, but it appears this is not the case. The memory usage increases to the point where it's killed by the OS. I think I must be missing something. Can you help me understand how to do this?

1. Does eland indeed do this automagically or not?
2. With pandas this can be achieved through the parameter `chunksize`. I've tried a lot of values, but this appears to do nothing, memory grows always in the same way.
3. The last line I get in the logs before the OS kills the process is the `REQUEST` request
```
[lq0jg ](https://swarmpit.soccrates.xyz/#/tasks/lq0jgu4ewxdcoqcauan7yr2nu?log=1) 2022-03-29 13:44:48,182 - INFO: POST https://es-coordinator:9200/_search [status:200 request:0.378s]
[lq0jg ](https://swarmpit.soccrates.xyz/#/tasks/lq0jgu4ewxdcoqcauan7yr2nu?log=1) 2022-03-29 13:44:49,169 - INFO: POST https://es-coordinator:9200/_search [status:200 request:0.158s]
[lq0jg ](https://swarmpit.soccrates.xyz/#/tasks/lq0jgu4ewxdcoqcauan7yr2nu?log=1) 2022-03-29 13:44:49,884 - INFO: POST https://es-coordinator:9200/_search [status:200 request:0.006s]
[lq0jg ](https://swarmpit.soccrates.xyz/#/tasks/lq0jgu4ewxdcoqcauan7yr2nu?log=1) 2022-03-29 13:44:49,887 - INFO: DELETE https://es-coordinator:9200/_pit [status:200 request:0.003s]
```

V1NAY8:
Closes #449 

We are dumping the entire ES index to pandas data frame and then converting it into csv. This might not hold for large datasets and cause unresponsiveness.

So, the fix is to write the data when we are fetching data as chunks from ES to csv file directly with `DEFAULT_SEARCH_SIZE = 5000`. 

In this way the data frame size will remain constant to that of five thousand rows every time when converting to CSV.

Also handled is the case where csv output is returned as string when no file path or buffer is provided.

@sethmlarson Please review and run jenkins

### DIFF
--- a/eland/operations.py
+++ b/eland/operations.py
@@ -1225,36 +1225,56 @@ class Operations:
         )
 
     def to_pandas(
-        self, query_compiler: "QueryCompiler", show_progress: bool = False
-    ) -> pd.DataFrame:
+        self,
+        query_compiler: "QueryCompiler",
+        show_progress: bool = False,
+        to_csv: bool = False,
+        **csv_kwargs: Union[bool, str],
+    ) -> Optional[Union[pd.DataFrame, str]]:
+
+        write_header: bool = True
+        _ = csv_kwargs.pop("mode", None)
+        _ = csv_kwargs.pop("header", None)
+
+        # Return a csv string if file path is not provided.
+        result = []
+        path_or_buf = csv_kwargs.pop("path_or_buf", None)
 
         df_list: List[pd.DataFrame] = []
         i = 0
+
         for df in self.search_yield_pandas_dataframes(query_compiler=query_compiler):
             if show_progress:
                 i = i + df.shape[0]
                 if i % DEFAULT_PROGRESS_REPORTING_NUM_ROWS == 0:
                     print(f"{datetime.now()}: read {i} rows")
-            df_list.append(df)
+
+            if to_csv:
+                result.append(
+                    df.to_csv(
+                        path_or_buf=path_or_buf,
+                        mode="w" if write_header else "a",
+                        header=True if write_header else False,
+                        **csv_kwargs,
+                    )
+                )
+                write_header = False
+            else:
+                df_list.append(df)
 
         if show_progress:
             print(f"{datetime.now()}: read {i} rows")
 
-        # pd.concat() can't handle an empty list
-        # because there aren't defined columns.
-        if not df_list:
-            return query_compiler._empty_pd_ef()
-        return pd.concat(df_list)
-
-    def to_csv(
-        self,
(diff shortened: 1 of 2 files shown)


eland/operations.py
```diff
@@ -1225,36 +1225,56 @@ def describe(self, query_compiler: "QueryCompiler") -> pd.DataFrame:
         )
 
     def to_pandas(
```
sethmlarson, line 1227: Instead of modifying `to_pandas()` what are your thoughts on modifying on `to_csv()` to chunk the dataframes with `self.search_yield_pandas_dataframes(...)` and then progressively appending them with `to_csv()`.

V1NAY8: Basically  we have to support `show_progress` and the chunkingpart. If we combine these two it becomes another `to_pandas` method. 

So duplication of code

V1NAY8: I mean only 4 lto 5 lines will be different between two methods. If i re implement it.

sethmlarson: My problem with the proposed implementation is it changes `to_pandas()` to do two completely different things based on the value of a parameter. Whenever you see this you likely actually want two different functions instead of one function filling two roles.

The above pain that you mention of "having to reimplement to_pandas" be hinting that we need a separate method (maybe `iterframes`?) that does everything inside of `to_pandas()` except it yields dataframes instead of stitching them together at the end. Then we can change `to_pandas()` and `to_csv()` to use the new `iterframes` method.

What do you think about this?

V1NAY8: This looks good. Ill implement. Never thought of that😶😅

falcorocks:
Hey I can't contribute much but thanks for opening this to close https://github.com/elastic/eland/issues/449 that I opened!

eland/operations.py
bartbroere: @V1NAY8 Thanks for working on this. It would really benefit my use case as well! Would you be willing to continue your work here? 

If you don't have the time right now, that's fine of course. In that case, I might build upon your changes and try to work in @sethmlarson's suggestions, if that's okay with you.

pquentin:
Merged as part of #579

PR #458, opened 2022-04-19
[ML] fixing XGBoost model import for xgboost>=1.6

benwtrent:
XGBoost >= 1.6 has removed the `booster` attribute from the underlying model `Booster` class.

This commit addresses this by adjusting the booster check to verify the the class has the attribute before checking that value.

Otherwise, we look at the top level model object `booster` attribute (which works fine for 1.6)

### DIFF
--- a/eland/ml/transformers/xgboost.py
+++ b/eland/ml/transformers/xgboost.py
@@ -125,8 +125,6 @@ class XGBoostForestTransformer(ModelTransformer):
 
         :return: A list of Tree objects
         """
-        self.check_model_booster()
-
         tree_table: pd.DataFrame = self._model.trees_to_dataframe()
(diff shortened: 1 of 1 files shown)


eland/ml/transformers/xgboost.py
```diff
@@ -155,17 +153,16 @@ def determine_target_type(self) -> str:
     def is_objective_supported(self) -> bool:
         return False
 
-    def check_model_booster(self) -> None:
+    @staticmethod
+    def check_model_booster(booster: Optional[str]) -> None:
```
sethmlarson, line 157: Since we're already calling this function in two places, thoughts on pulling the `booster` string grabbing logic into this function and passing the model as a parameter?

PR #459, opened 2022-04-19
[ML] Improve NLP model import by using nicely defined types

benwtrent:
This adds some more definite types for our NLP tasks and tokenization configurations.

This is the first step in allowing users to more easily import their own transformer models via something other than hugging face.

### DIFF
--- a/eland/ml/pytorch/_pytorch_model.py
+++ b/eland/ml/pytorch/_pytorch_model.py
@@ -19,7 +19,17 @@ import base64
 import json
 import math
 import os
-from typing import TYPE_CHECKING, Any, Iterable, List, Mapping, Set, Tuple, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Iterable,
+    List,
+    Mapping,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+)
 
 from tqdm.auto import tqdm  # type: ignore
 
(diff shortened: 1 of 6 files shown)


eland/ml/pytorch/_pytorch_model.py
```diff
@@ -88,12 +89,14 @@ def model_file_chunk_generator() -> Iterable[str]:
     def import_model(
         self,
         model_path: str,
-        config_path: str,
```
sethmlarson, line 91: This would technically be a break as `PyTorchModel` is a public facing API. We could make the `import_model` method keyword only, have both `config_path` and `config` as parameters, and emit a `DeprecationWarning` if `config_path` is used as a way to use `config` in the future?

eland/ml/pytorch/nlp_ml_model.py
```diff
@@ -0,0 +1,248 @@
...
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+import typing as t
+
+
+class NlpTokenizationConfig:
```
sethmlarson, line 21: Would recommend making all of these classes `__init__` parameters keyword only via `__init__(self, *, ...)` so they can grow more easily.

sethmlarson (commented):
Left some comments for you:

eland/ml/pytorch/_pytorch_model.py
benwtrent: Do we have any rules around tech-preview features? The APIs being used in Elasticsearch are techpreview and subject to change. Seems like the client objects interacting with them should have the same bwc guarantees no?

eland/ml/pytorch/nlp_ml_model.py
```diff
@@ -0,0 +1,232 @@
...
+        classification_labels: t.Union[t.List[str], t.Tuple[str, ...]],
+        tokenization: NlpTokenizationConfig,
+        results_field: t.Optional[str] = None,
+        num_top_classes: t.Optional[int] = None,
+    ):
+        super().__init__(configuration_type="text_classification")
+        self.results_field = results_field
+        self.num_top_classes = num_top_classes
+        self.tokenization = tokenization
+        self.results_field = results_field
```
davidkyle, line 123:
```suggestion
```

Appears twice

eland/ml/pytorch/transformers.py
```diff
@@ -45,6 +59,13 @@
     "text_embedding",
     "zero_shot_classification",
 }
+TASK_TYPE_TO_INFERENCE_CONFIG = {
+    "fill_mask": FillMaskInferenceOptions,
```
davidkyle, line 63: You've defined `PassThroughInferenceOptions` may as well use it.

davidkyle (approved):
LGTM

PR #460, opened 2022-04-21
Align dependencies between requirement files and setup.py

dolaru:
Dependencies weren't matching between `setup.py` and `requirements.txt`/`requirements-dev.txt`. 

This PR aligns them to match, and locks the dependency version to the current major where a version was previously specified.

Note: 
`scikit-learn` was locked to versions lower than `1.0.0` in `setup.py`, but it was open-ended on the high side in `requirements-dev.txt`. At the time this PR was raised, the latest version is `1.0.2`, which is why I locked the max version `<2`.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -10,7 +10,7 @@ tqdm<5
 #
 # Extras
 #
-scikit-learn>=0.22.1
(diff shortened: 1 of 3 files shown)


sethmlarson (approved):
LGTM

PR #461, opened 2022-04-21
Hub model import script improvements

dolaru:
### Better logging
Switched from `print` statements to `logging` for a cleaner and more informative output - timestamps and log level are shown. The logging is now a bit more verbose, but it will help users to better understand what the script is doing.

### Add support for ES authentication using username/password or api key
Instead of being limited to passing credentials in the URL, there are now 2 additional methods:
- username/password using `--es-username` and `--es-password`
- API key using `--es-api-key`

Credentials can also be specified as environment variables with `ES_USERNAME`/`ES_PASSWORD` or `ES_API_KEY`

### Graceful handling of missing PyTorch requirements
In order to use the `eland_import_hub_model` script, PyTorch extras are required to be installed. If the user does not have the required packages installed, a helpful message is logged with a hint to install `eland[pytorch]` with `pip`.

### Graceful handling of already existing trained model
If a trained model with the same ID as the one we're trying to import already exists, and `--clear-previous` was not specified, we now log a clearer message about why the script can't proceed along with a hint to use the `--clear-previous` flag. 

Prior to this change, we were letting the API exception seep through and the user was faced with a stack trace.

### `tqdm` added to main dependencies
If the user doesn't have `eland[pytorch]` extras installed, the first module to be reported as missing is `tqdm`. Since this module is [used in eland codebase](https://github.com/elastic/eland/blob/8294224e3436a595d6bf0834c48e3347907cfe4c/eland/ml/pytorch/_pytorch_model.py#L24) directly, it makes sense to me to have it as part of the main set of requirements.

### Nit: Set tqdm unit to `parts` in `_pytorch_model.put_model`
The default unit is `it`, but `parts` better describes what the progress bar is tracking - uploading trained model definition parts.

### DIFF
--- a/bin/eland_import_hub_model
+++ b/bin/eland_import_hub_model
@@ -24,39 +24,58 @@ uploading to Elasticsearch. This will also check that the task type is supported
 as well as the model and tokenizer types. All necessary configuration is
 uploaded along with the model.
 """
-
 import argparse
+import logging
+import os
+import sys
 import tempfile
+import textwrap
 
-import elasticsearch
 from elastic_transport.client_utils import DEFAULT
-
-from eland.ml.pytorch import PyTorchModel
-from eland.ml.pytorch.transformers import SUPPORTED_TASK_TYPES, TransformerModel
+from elasticsearch import AuthenticationException, Elasticsearch
 
 MODEL_HUB_URL = "https://huggingface.co"
 
 
-def main():
-    parser = argparse.ArgumentParser(prog="upload_hub_model")
+def get_arg_parser():
+    parser = argparse.ArgumentParser()
     parser.add_argument(
         "--url",
         required=True,
-        help="An Elasticsearch connection URL, e.g. http://user:secret@localhost:9200",
+        default=os.environ.get("ES_URL"),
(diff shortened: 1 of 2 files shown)


benwtrent (approved):
lgtm

bin/eland_import_hub_model
```diff
@@ -93,36 +112,100 @@ def main():
...
-        if args.clear_previous:
-            print(f"Stopping previous deployment and deleting model: {ptm.model_id}")
-            ptm.stop()
-            ptm.delete()
-        print(f"Importing model: {ptm.model_id}")
-        ptm.import_model(model_path, config_path, vocab_path)
-
-    # start the deployed model
+        ptm = PyTorchModel(es, args.es_model_id if args.es_model_id else tm.elasticsearch_model_id())
+        model_exists = es.ml.get_trained_models(model_id=ptm.model_id, ignore=404).meta.status == 200
```
sethmlarson, line 181: Should use `es.options(ignore_status=404).ml.get_trained_models(...)` instead of `ignore=404` which is deprecated.

sethmlarson (commented):
Looks good, one comment on `ignore=` being deprecated:

sethmlarson (approved):

PR #462, opened 2022-04-27
Add support for Cloud ID to hub model import script

dolaru:
The Cloud ID simplifies sending data to a cluster on Elastic Cloud.

With this change, the user will have the option specify a Cloud ID using the `--cloud-id` argument as an alternative to an Elasticsearch URL (`--url` argument).

`--cloud-id` and `--url` are mutually exclusive arguments.


sethmlarson (approved):
LGTM, thanks!

PR #465, opened 2022-05-11
[ML] fixes decision tree classifier upload to account for probabilities

benwtrent:
This switches our sklearn.DecisionTreeClassifier serialization logic to account for multi-valued leaves in the tree.

The key difference between our inference and DecisionTreeClassifier, is that we run a `softMax` over the leaf where sklearn simply normalizes the results.

This means that our "probabilities" returned will be different than sklearn.

### DIFF
--- a/eland/ml/ml_model.py
+++ b/eland/ml/ml_model.py
@@ -269,6 +269,9 @@ class MLModel:
 
         model: An instance of a supported python model. We support the following model types:
             - sklearn.tree.DecisionTreeClassifier
+                - NOTE: When calculating the probabilities of a given classification label, Elasticsearch utilizes
+                        softMax. SKLearn instead normalizes the results. This may result in different probabilities
+                        but the overall predictions will still align.
             - sklearn.tree.DecisionTreeRegressor
             - sklearn.ensemble.RandomForestRegressor
             - sklearn.ensemble.RandomForestClassifier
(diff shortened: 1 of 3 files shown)


eland/ml/ml_model.py
```diff
@@ -269,6 +269,9 @@ def import_model(
 
         model: An instance of a supported python model. We support the following model types:
             - sklearn.tree.DecisionTreeClassifier
+                - NOTE: When calculating the probabilities of a given classification label, Elasticsearch utilizes
+                        softMax. SKLearn instead normalizes the results. This may result in different probabilities
+                        but the overall predictions will still align.
```
benwtrent, line 274: @tveasey can you think of a way of adjusting the raw leaf values so that when we take the softmax it ends up being the same as normalizing the raw values?

benwtrent: Well, `log` could work if I just choose some large negative number for when `n` is 0. I will experiment.

eland/ml/transformers/sklearn.py
```diff
@@ -86,8 +86,12 @@ def build_tree_node(
             ):  # classification requires more than one value, so assume regression
                 leaf_value = [float(value[0][0])]
             else:
-                # the classification value, which is the index of the largest value
-                leaf_value = [float(np.argmax(value))]
+                # the classification value
+                # DecisionTreeClassifiers simply use normalize (dividing predicted values by sum)
+                # We use softMax, to get our probabilities as close as possible, store log value here
+                leaf_value = [
+                    -10000000 if n == 0 else math.log(float(n)) for n in value[0]
```
tveasey, line 93: Feels like we should be careful about negative values. Shouldn't crop up but we can be defensive. Either error out or treat as zero, e.g.
```suggestion
                    -10000000 if n <= 0 else math.log(float(n)) for n in value[0]
```

tveasey (approved):
Might be worthwhile being defensive.

PR #473, opened 2022-06-21
Freeze the traced PyTorch model

davidkyle:
Freezing PyTorch/TorchScript models is an optimisation of the internal format that strips attributes and inlines constant values. A frozen model should only be used for inference. 

Benchmarking with 2 commonly used NLP models (https://huggingface.co/sentence-transformers/multi-qa-MiniLM-L6-cos-v1 and https://huggingface.co/sentence-transformers/all-distilroberta-v1) shows there is a small performance gain of < 1% when using frozen models. The gain is small there is no downside to freezing models that will only be used for inference.


droberts195 (approved):
LGTM

benwtrent (approved):

sethmlarson (approved):

PR #475, opened 2022-06-22
[ML] adds new auto task type that attempts to automatically determine NLP task type from model config

benwtrent:
For many model types, we don't need to require the task requested. We can infer the task type based on the model configuration and architecture. 

This commit makes the `task-type` parameter optional for the model up load script and adds logic for auto-detecting the task type based on the 🤗 model.

### DIFF
--- a/bin/eland_import_hub_model
+++ b/bin/eland_import_hub_model
@@ -196,7 +196,7 @@ if __name__ == "__main__":
         try:
             tm = TransformerModel(args.hub_model_id, args.task_type, args.quantize)
             model_path, config, vocab_path = tm.save(tmp_dir)
-        except TransformerModel as err:
+        except TaskTypeError as err:
             logger.error(f"Failed to get model for task type, please provide valid task type via '--task-type' parameter. Caused by {err}")
             exit(1)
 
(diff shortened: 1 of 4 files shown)


bin/eland_import_hub_model
```diff
@@ -187,8 +193,12 @@ if __name__ == "__main__":
     with tempfile.TemporaryDirectory() as tmp_dir:
         logger.info(f"Loading HuggingFace transformer tokenizer and model '{args.hub_model_id}'")
 
-        tm = TransformerModel(args.hub_model_id, args.task_type, args.quantize)
-        model_path, config, vocab_path = tm.save(tmp_dir)
+        try:
+            tm = TransformerModel(args.hub_model_id, args.task_type, args.quantize)
+            model_path, config, vocab_path = tm.save(tmp_dir)
+        except TransformerModel as err:
```
davidkyle, line 199:
```suggestion
        except TaskTypeError as err:
```

davidkyle (approved):
NICE

bin/eland_import_hub_model
benwtrent: 🤦

PR #478, opened 2022-06-28
Add more exclusions to the dockerignore file

davidkyle:
Makes the `.dockerignore` file more restrictive by matching what is in `.gitignore`. Of particular note is the exclusion of virtual environments can be large


droberts195 (approved):
LGTM

PR #479, opened 2022-06-30
Restrict PyTorch version not to be more advanced than that used in Elasticsearch 

davidkyle:
PyTorch recently released [version 1.12](https://github.com/pytorch/pytorch/releases/tag/v1.12.0) which contains breaking changes to torchscript models making them incompatible with earlier versions. 

The Elastic stack uses version 1.11 of the PyTorch C++ library (libtorch), a model traced to torchscript format using a v1.12 Python PyTorch install cannot be evaluated in libtorch v1.11. 

Anyone using Eland to import a model into Elastic today would pull the latest version of PyTorch and find the model cannot be evaluated in Elasticsearch. Opening the model errors with the message: 

`Attempted to read a PyTorch file with version 10, but the maximum supported version for reading is 9. Your PyTorch installation may be too old.` 

This change to the requirements prevents Eland using v1.12 of PyTorch.

HuggingFace transformers has also pinned the torch requirement to `<1.12` in https://github.com/huggingface/transformers/commit/5a3d0cbdda73afafebbb8516901e622242a67863

The `sentence-transformers` and `transformers` versions have also been tighten to a range known to be compatible with PyTorch 1.11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,10 @@ $ conda install -c conda-forge eland
 ### Compatibility
 
 - Supports Python 3.7+ and Pandas 1.3
-- Supports Elasticsearch clusters that are 7.11+, recommended 7.14 or later for all features to work.
-  Make sure your Eland major version matches the major version of your Elasticsearch cluster.
+- Supports Elasticsearch clusters that are 7.11+, recommended 8.3 or later for all features to work.
+  If you are using the NLP with PyTorch feature make sure your Eland minor version matches the minor 
+  version of your Elasticsearch cluster. For all other features it is sufficient for the major versions
+  to match.
 
 ### Prerequisites
 
--- a/eland/ml/pytorch/_pytorch_model.py
+++ b/eland/ml/pytorch/_pytorch_model.py
@@ -125,9 +125,7 @@ class PyTorchModel:
         docs: List[Mapping[str, str]],
         timeout: str = DEFAULT_TIMEOUT,
     ) -> Any:
-        return self._client.options(
-            request_timeout=60
-        ).ml.infer_trained_model(
+        return self._client.options(request_timeout=60).ml.infer_trained_model(
             model_id=self.model_id,
             timeout=timeout,
             docs=docs,
--- a/eland/ml/pytorch/_pytorch_model.py
+++ b/eland/ml/pytorch/_pytorch_model.py
@@ -127,7 +127,7 @@ class PyTorchModel:
     ) -> Any:
         return self._client.options(
             request_timeout=60
-        ).ml.infer_trained_model_deployment(
+        ).ml.infer_trained_model(
             model_id=self.model_id,
             timeout=timeout,
             docs=docs,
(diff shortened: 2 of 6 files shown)


droberts195:
It sounds like the `README` needs updating too.  Currently it says:

> * Supports Elasticsearch clusters that are 7.11+, recommended 7.14 or later for all features to work. Make sure your Eland major version matches the major version of your Elasticsearch cluster.

But it sounds like actually the requirement is that your Eland distribution uses the same PyTorch version as the ML C++ code, and to avoid having some really complicated compatibility matrix the easiest way to express that is to say Eland and Elasticsearch must be of the same minor version (any patch level within that minor).

benwtrent (approved):

davidkyle:
The CI failure was due to a change in the Python client after `_infer` API was renamed in the Elasticsearch 8.3 release.

https://github.com/elastic/elasticsearch-py/commit/0e16cd6930db026cfc5dc35469682af0300d022e

As the the `_infer` API is in tech preview and the old name is supported but deprecated it is ok to rename but it means a 8.3 version of Eland must be used with an 8.3 Elasticsearch cluster. 

Given that upgrading PyTorch has already enforced this requirement and may happen again on a later upgrade I've updated the README to state that if using the PyTorch features the minor versions of Elasticsearch and Eland must be the same.

davidkyle:
I pushed c97d641bebb41540c9917467278bb83d67b11e2a which bumps the required Elasticsearch Python client to version 8.3.

Eland does not make any claims for compatibility with older versions of the Python client and the latest is always preferred

droberts195:
> we will make it backwards compatible with earlier 8.x versions and future compatible with later 8.x versions

Is it the case that if you trace a model using PyTorch 1.11 then PyTorch 1.9 cannot understand it?  If that is the case then it's going to be very hard to make Eland compatible with older stack versions because it will have to dynamically get the appropriate version of PyTorch after finding out what the server version is.

davidkyle:
> Is it the case that if you trace a model using PyTorch 1.11 then PyTorch 1.9 cannot understand it?

Anecdotally we know that models traced in PyTorch 1.11 can run in libtorch 1.9 as that is what many people have been doing. But it's worth verifying so I spun up a cloud cluster of Elasticsearch version 8.0.1 and uploaded a NER model using the changes in this PR (with PyTorch pinned to 1.11). Evaluating the model did work 😁 so PyTorch 1.11 is backward compatible with 1.9, the breaking change is in 1.12.

droberts195:
> the breaking change is in 1.12

But then we're setting ourselves up for a problem the next time we upgrade PyTorch if we say Eland 8.3 and above is going to be compatible with all Elasticsearch 8.x versions.

We could say Eland 8.3 is compatible with all Elasticsearch 8.x versions and then never release another version of Eland.  Or we could never upgrade the version of PyTorch used within Eland again in 8.x, so that by 8.last Elasticsearch might be on PyTorch 1.17 but Eland would still be on PyTorch 1.11.

But neither of those options sounds ideal. We could be forced into doing a quick upgrade by a major security bug for example.

This is why I think it would be better to define a compatibility policy for Eland and Elasticsearch that accounts for the possibility of breaking changes in PyTorch.

droberts195 (approved):
LGTM

mrklaney:
Whaaaaattttttttt!   Do you know how many weeks I have been trying to get my model not to crash?!

PR #486, opened 2022-07-11
[ML] add text_similarity task support

benwtrent:
Adds text_similarity task support. This is a cross-encoder transformer task where both sequences are given to the transformer at once. 

According to 🤗 (or at least how the cross-encoder models are concerned) this is a sequence classification task with just one classification "label". But really, it isn't labeled at all and is more akin to a regression model.

related: https://github.com/elastic/elasticsearch/pull/88439

### DIFF
--- a/eland/ml/pytorch/nlp_ml_model.py
+++ b/eland/ml/pytorch/nlp_ml_model.py
@@ -218,6 +218,20 @@ class QuestionAnsweringInferenceOptions(InferenceConfig):
         self.num_top_classes = num_top_classes
 
 
+class ReRankingInferenceOptions(InferenceConfig):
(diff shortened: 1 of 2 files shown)


dimitris-athanasiou (approved):
LGTM

dzlab:
@benwtrent how is this can be used in eland directly? can you point to examples the same way with elasticsearch API here https://github.com/elastic/elasticsearch/pull/88439

dimitris-athanasiou:
@dzlab This is just about being able to upload a `text_similarity` model to elasticsearch. After that is done then you can interact with the model like in https://github.com/elastic/elasticsearch/pull/88439.

PR #487, opened 2022-07-12
[DOCS] Adds important note about PyTorch version compatibility

szabosteve:
## Overview

The latest version of PyTorch 1.12 is not compatible with ES 8.3. Users need to use 1.11.0 or earlier to import a model via eland. This PR adds an important note about it to the docs and the README.


davidkyle (approved):
LGTM

PR #505, opened 2022-11-24
Handle lint failures

davidkyle:


eland/arithmetics.py
```diff
@@ -33,6 +33,7 @@ class ArithmeticObject(ABC):
     def value(self) -> str:
         pass
 
+    @property
```
davidkyle, line 36: The sub class implementations have the `@property` decorator, seems like an oversight not to have it here

eland/common.py
```diff
@@ -345,5 +345,5 @@ def es_version(es_client: Elasticsearch) -> Tuple[int, int, int]:
             )
 
     else:
-        eland_es_version = es_client._eland_es_version  # type: ignore
```
davidkyle, line 348: Before I removed this the linter was failing with 

```
nox > Session lint aborted: 
eland/common.py:348: error: Unused "type: ignore" comment.
```

But when I tested the change on another machine with different versions of the Python, mypy and nox installed removing this comment caused `nox -s format` to fail. 

I'm a bit confused about what to do here, do you have any suggestions please @sethmlarson

eland/arithmetics.py
sethmlarson: Ah yeah, likely an oversight. Thanks!

eland/common.py
sethmlarson: It's because Mypy has been upgraded and the other machine is likely using a cached copy of the old Mypy. Try running with `nox -rs format`, the `-r` will recreate the environment and force the new version to be installed.

This removal LGTM.

sethmlarson (approved):
Thank you! :tada:

eland/common.py
davidkyle: I took the nuclear option of blowing away the `.nox` directory and indeed the format problem went away after that.

PR #506, opened 2022-11-30
Include pitfall of `--start` in the README

droberts195:
Users who follow the Eland README as a guide to importing models can easily end up seeing inexplicably poor performance due to unknowingly running the model with one allocation and one thread per allocation.

This change spells out the effect of `--start` and links to alternatives that allow better use of available hardware.


davidkyle (approved):
LGTM

PR #509, opened 2022-12-19
[ML] Export ML model as sklearn Pipeline

### LINKED ISSUE #503: Export gradient boosting models trained in Elastic stack
In Elastic Stack we can train supervised learning models to create an ensemble of gradient-boosting trees. This issue suggests extending the codebase of eland to be able to download these models on a local machine as an instance of [`sklearn.Pipeline`](https://scikit-learn.org/stable/modules/compose.html):
```python
es_model = eland.ml.MLModel(es_client=es_client, model_id=model_id)
pipeline = es_model.export_model()
pipeline.fit(X)
```

## Why?

Being able to treat the ES models as `sklearn` pipelines will allow data scientists to use their usual tools to analyze, explore, or debug the models, prove unbiasedness, etc. 

## How?

We can convert the model JSON definition into a subclass of  `GradientBoostingClassifier` or `GradientBoostingRegressor` of `sklearn`. The encoding preprocessor can be represented as `ColumnTransformer`.  Once we have these individual components, we can use `sklearn.Pipeline` to produce a single object that can be evaluated by passing the original data frame.

We will be able to run the functionality similarly and obtain numerically similar results:
```python
es_prediction = es_model.predict_proba(X, feature_names_in=pipeline["preprocessor"].get_feature_names_out())
sklearn_prediction = pipeline.predic_proba(X, feature_names_in=pipeline["preprocessor"].get_feature_names_out())

numpy.tsting.assert_allclose(es_prediction, sklearn_prediction)
```

## Limitations

Since we have to instantiate the model locally, we will be limited by the memory size and computational capacity of a laptop/desktop. We will document this limitation, although most models should be able to fit into 8-16GB of memory available on modern PCs.

valeriy42:
Closes #503 

Note: I also had to fix the Sphinx version to 5.3.0 since starting from 6.0, Sphinx suffers from a [TypeError bug](https://github.com/sphinx-doc/sphinx/issues/11094), which causes a CI failure.


eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,434 @@
...
+            - elasticsearch-py instance
+        model_id : str
+            The unique identifier of the trained inference model in Elasticsearch.
+
+        Raises
+        ------
+        NotImplementedError
+            Multi-class classification is not supported at the moment.
+        ValueError
+            The classifier should be defined for at least 2 classes.
```
tveasey, line 203: Can this also raise KeyError if names were to change.

tveasey (commented):
Aside from some minor stuff there was one thing which looks suspicious to me. I'd like to do a second read through as well. Regarding comments on error handling for map access with special names, there maybe isn't much more we can do, but perhaps we can just think about if there is anything extra to add to exceptions that'd help us debug an issue.

eland/ml/exporters/_sklearn_deserializers.py
```diff
@@ -0,0 +1,210 @@
...
+            values,
+            right_index,
+            depth + 1,
+        )
+        left_weight = node_sample_weight[left_index]
+        right_weight = node_sample_weight[right_index]
+
+        v = (
+            left_weight * values[left_index, :] + right_weight * values[right_index, :]
+        ) / (left_weight + right_weight)
```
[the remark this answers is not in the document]
valeriy42: 👍 good catch.

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,434 @@
...
+    HuberLossFunction,
+    LeastSquaresError,
+)
+from sklearn.tree import DecisionTreeClassifier, DecisionTreeRegressor
+from sklearn.utils.validation import check_array
+
+from eland.common import ensure_es_client
+from eland.ml.common import TYPE_CLASSIFICATION, TYPE_REGRESSION
+
+from ._sklearn_deserializers import Tree
```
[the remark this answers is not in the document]
valeriy42: The relative import paths are used everywhere else in the code, so I adhere to the convention.

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,434 @@
...
+        Raises
+        ------
+        ValueError
+            The model is expected to be trained in Elastic Stack. Models initially imported
+            from xgboost, lgbm, or sklearn are not supported.
+        """
+        self.es_client: Elasticsearch = ensure_es_client(es_client)
+        self.model_id = model_id
+
+        self._trained_model_result = self.es_client.ml.get_trained_models(
```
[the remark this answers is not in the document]
valeriy42: I added raising `RuntimeError` if I fail to get the trained model definition. It's similar behavior as in `predict()`.

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,434 @@
...
+
+    def _get_feature_names_in_(
+        self, preprocessors, feature_names, field_names
+    ) -> Tuple[List[str], Set[str]]:
+        input_field_names = set()
+        for preprocessor in preprocessors:
+            if "target_mean_encoding" in preprocessor:
+                feature_name = preprocessor["target_mean_encoding"]["feature_name"]
+                if feature_name in feature_names:
+                    input_field_names.add(preprocessor["target_mean_encoding"]["field"])
```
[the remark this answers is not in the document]
valeriy42: Added a subroutine

eland/ml/exporters/es_gb_models.py
valeriy42: If the JSON schema of the trained model definition changes, it would also generate the KeyError. However, DFA is GA, so the schema is not supposed to change so quickly, and [Elasticsearch Python clients are only backward compatible](https://elasticsearch-py.readthedocs.io/en/v8.6.0/#compatibility) within the same major.

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,449 @@
...
+        )
+
+        feature_names_map = {name: i for i, name in enumerate(self.feature_names_in_)}
+
+        trained_models = self._definition["trained_model"]["ensemble"]["trained_models"]
+        self._trees = []
+        for trained_model in trained_models:
+            self._trees.append(Tree(trained_model["tree"], feature_names_map))
+
+        self.n_estimators = (
```
sethmlarson, line 109: This line got formatted weirdly, perhaps move the comment above the line and reformat?

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,449 @@
...
+
+    @property
+    def _n_outputs(self) -> int:
+        return self._trees[0].n_outputs
+
+    @property
+    def _definition(self) -> Mapping[Union[str, int], Any]:
+        return self._trained_model_result["trained_model_configs"][0]["definition"]
+
+    def _get_feature_names_in_(
```
sethmlarson, line 141: Since this doesn't use `self` we can make this a `@staticmethod`.

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,449 @@
...
+        trained_models = self._definition["trained_model"]["ensemble"]["trained_models"]
+        self._trees = []
+        for trained_model in trained_models:
+            self._trees.append(Tree(trained_model["tree"], feature_names_map))
+
+        self.n_estimators = (
+            len(trained_models) - 1
+        )  # 0's tree is the constant estimator
+
+    def _initialize_estimators(self, DecisionTreeType) -> None:
```
sethmlarson, line 113: nit: `decision_tree_type` over `DecisionTreeType`.

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,449 @@
...
+    @property
+    @abstractmethod
+    def analysis_type(self):
+        """
+        Type of the data frame analysis. It can be classification or regression.
+        """
+        pass
+
+    @abstractmethod
+    def _initialize_init_(self) -> Union[DummyClassifier, DummyRegressor]:
```
sethmlarson, line 190: Is this needed on the ABC? Looks like it's only used in implementations.

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,449 @@
...
+
+        if self.n_classes_ == 2:
+            self._loss = BinomialDeviance(self.n_classes_)
+            # self.n_outputs = 1
+        elif self.n_classes_ > 2:
+            raise NotImplementedError("Only binary classification is implemented.")
+        else:
+            raise ValueError(f"At least 2 classes required. got {self.n_classes_}.")
+
+        self.init_ = self._initialize_init_()
```
sethmlarson, line 252: Is `self.init_` used anywhere? It looks like we're returning an estimator or something like this, if it is needed we should use a better name.

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,449 @@
+#  Licensed to Elasticsearch B.V. under one or more contributor
```
sethmlarson, line 1: Is this module supported to be public, or can we make it private and expose the classes via `eland.ml.ml_model`?

eland/ml/ml_model.py
```diff
@@ -424,6 +435,63 @@ def exists_model(self) -> bool:
...
+            field = p[encoding_type]["field"]
+            if encoding_type == "frequency_encoding":
+                transform = FrequencyEncoder(p)
+                transformers.append((f"{field}_{encoding_type}", transform, field))
+            elif encoding_type == "target_mean_encoding":
+                transform = TargetMeanEncoder(p)
+                transformers.append((f"{field}_{encoding_type}", transform, field))
+            elif encoding_type == "one_hot_encoding":
+                transform = OneHotEncoder(p)
+                transformers.append((f"{field}_{encoding_type}", transform, [field]))
```
sethmlarson, line 484: We need an error case here if the `encoding_type` is something else?

eland/ml/ml_model.py
```diff
@@ -424,6 +435,63 @@ def exists_model(self) -> bool:
...
+        else:
+            raise NotImplementedError
+
+        if "preprocessors" in model._definition:
+            preprocessors = model._definition["preprocessors"]
+        else:
+            preprocessors = []
+        transformers = []
+        for p in preprocessors:
+            encoding_type = list(p.keys())[0]
```
sethmlarson, line 474: Since this is assuming a one-key mapping can we add an assertation of `len(p) == 1`?

eland/ml/ml_model.py
```diff
@@ -424,6 +435,63 @@ def exists_model(self) -> bool:
...
+        if self.model_type == TYPE_CLASSIFICATION:
+            model = ESGradientBoostingClassifier(
+                es_client=self._client, model_id=self._model_id
+            )
+        elif self.model_type == TYPE_REGRESSION:
+            model = ESGradientBoostingRegressor(
+                es_client=self._client, model_id=self._model_id
+            )
+        else:
+            raise NotImplementedError
```
sethmlarson, line 466: Can we have a more descriptive error message here?

sethmlarson (changes requested):
Thanks for this, looks mostly good! A few thoughts for you below.

Also there's a handful of places where we need to remove the underscore suffix on names, I didn't comment on all of them. Searching for `_ =` might be the best method.

eland/ml/exporters/_sklearn_deserializers.py
```diff
@@ -0,0 +1,210 @@
...
+import sklearn
+from sklearn.preprocessing import FunctionTransformer
+
+
+class Tree:
+    """Wrapper to create sklearn Tree objects from Elastic ML tree
+    description in JSON format.
+    """
+
+    def __init__(self, json_tree, feature_names_map):
```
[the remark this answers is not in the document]
valeriy42: Type hinting added.

eland/ml/exporters/_sklearn_deserializers.py
```diff
@@ -0,0 +1,210 @@
...
+        children_right = np.ones((node_count,), dtype=int) * TREE_LEAF
+        feature = np.ones((node_count,), dtype=int) * -2
+        threshold = np.ones((node_count,), dtype=float) * -2
+        impurity = np.zeros((node_count,), dtype=float)
+        # value works only for regression and binary classification
+        value = np.zeros((node_count, 1, 1), dtype="<f8")
+        n_node_samples = np.zeros((node_count,), dtype=int)
+
+        # parse values from the JSON tree
+        feature_names = json_tree["feature_names"]
```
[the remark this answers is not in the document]
valeriy42: I created `ModelDefinitionKeyError` that gives additional information. Unfortunately, returning the list of available keys within the error is not straightforward. Such a change would make the code less readable.

eland/ml/exporters/_sklearn_deserializers.py
```diff
@@ -0,0 +1,212 @@
...
+    """Wrapper to create sklearn Tree objects from Elastic ML tree
+    description in JSON format.
+    """
+
+    def __init__(
+        self,
+        json_tree: Mapping[Union[str, int], Any],
+        feature_names_map: Dict[str, int],
+    ):
+        TREE_LEAF = -1
```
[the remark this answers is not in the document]
valeriy42: renamed

eland/ml/exporters/_sklearn_deserializers.py
```diff
@@ -0,0 +1,212 @@
...
+                ("weighted_n_node_samples", "<f8"),
+            ],
+        )
+        state = {
+            "max_depth": self.max_depth,
+            "node_count": node_count,
+            "nodes": node_state,
+            "values": value,
+        }
+        self.tree_.__setstate__(state)
```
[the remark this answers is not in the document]
valeriy42: Sometimes, I had to keep underscores to adhere to the sklearn naming conventions. In this case, it works without _, so I removed it.

eland/ml/exporters/_sklearn_deserializers.py
```diff
@@ -0,0 +1,212 @@
...
+
+        super().__init__(func=func, feature_names_out=feature_names_out)
+
+
+class FrequencyEncoder(FunctionTransformer):
+    """FunctionTransformer implementation of the frequency encoder, which is
+    deserialized from the Elastic ML preprocessor description in JSON format.
+    """
+
+    def __init__(self, preprocessor: Mapping[Union[str, int], Any]):
```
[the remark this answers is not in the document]
valeriy42: You are right. I can get away with using `Dict[str, Any]` here. Fixed it and at other places.

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,449 @@
...
+        if "metadata" not in self._trained_model_result["trained_model_configs"][0]:
+            raise ValueError(
+                "Error initializing sklearn classifier. Incorrect prior class probability. "
+                + "Note: only export of models trained in the Elastic Stack is supported."
+            )
+
+        preprocessors = []
+        if "preprocessors" in self._definition:
+            preprocessors = self._definition["preprocessors"]
+        self.feature_names_in_, self.input_field_names = self._get_feature_names_in_(
```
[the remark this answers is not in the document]
valeriy42: `feature_names_in_` is [an sklearn convention](https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep007/proposal.html).

eland/ml/exporters/es_gb_models.py
```diff
@@ -0,0 +1,449 @@
...
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+from abc import ABC, abstractmethod
+from typing import Any, List, Literal, Mapping, Optional, Set, Tuple, Union
+
+import numpy as np
+import scipy as sp
```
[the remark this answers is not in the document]
valeriy42: Thank you for pointing this out. I am using a single function, so I can just write it out explicitly to simplify the dependencies.

eland/ml/exporters/es_gb_models.py
valeriy42: fixed

eland/ml/exporters/es_gb_models.py
valeriy42: You are right. I removed both abstract method definitions.

eland/ml/exporters/es_gb_models.py
valeriy42: I am not sure what you mean here. Users can use the ESGradientBoosting* objects directly, accessing them from `pipeline['es_model']`. Can you elaborate?

eland/ml/ml_model.py
valeriy42: Sure :-)

eland/ml/ml_model.py
valeriy42: I added the handling of unexpected `encoding_type`.

eland/ml/ml_model.py
valeriy42: Assertion added

eland/ml/exporters/es_gb_models.py
valeriy42: `self.init_`  is a part of `sklearn` estimator definition. It is usually initialized when `fit()` is called. In our case, we don't use `fit()` since models are trained within Elasticsearch. Hence, I need to initialize it manually. `sklearn` is then uses `self.init_` for predictions, etc.

eland/ml/exporters/es_gb_models.py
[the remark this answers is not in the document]
tveasey: I think what triggered this comment was we import eland.common above, i.e. there seemed some inconsistency, but I can see advantage of importing from . directly and was only concerned that we followed repo conventions.

eland/ml/exporters/es_gb_models.py
tveasey: I guess I wonder if we should also document that. But take your point it is kind of an exceptional exception.

tveasey (approved):
Nice functionality! Thanks for iterating.

eland/ml/exporters/_sklearn_deserializers.py
```diff
@@ -0,0 +1,210 @@
...
+            children_left=children_left,
+            children_right=children_right,
+            node_sample_weight=weighted_n_node_samples,
+            values=value,
+            node_index=0,
+        )
+        self.n_outputs = value.shape[-1]
+
+        # initialize the sklearn tree
+        self.tree_ = sklearn.tree._tree.Tree(
```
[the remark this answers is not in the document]
tveasey: I guess if other packages rely on them then they've kind of leaked. My main concern is we'd might hit problems with sklearn versions. It might be worth adding your comment here as a code comment. Long term I guess we may have version dependent code here as well :(.

valeriy42:
Thank you for your review, @sethmlarson. I incorporated your feedback. Can you please take another look?

PR #512, opened 2023-01-23
Update to latest ES versions and fix unit tests

valeriy42:
Update the test matrix to the latest Elasticsearch versions and fix the broken unit tests on the CI.


docs/requirements-docs.txt
```diff
@@ -1,10 +1,11 @@
 elasticsearch>=7.7
-pandas>=1.2.0
-matplotlib
+pandas==1.5.0
```
droberts195, line 2: Please adjust the README too - that says 1.3+

noxfile.py
```diff
@@ -100,8 +100,8 @@ def lint(session):
             session.error("\n" + "\n".join(sorted(set(errors))))
 
 
-@nox.session(python=["3.7", "3.8", "3.9", "3.10"])
-@nox.parametrize("pandas_version", ["1.2.0", "1.3.0"])
+@nox.session(python=["3.8", "3.9", "3.10"])
```
droberts195, line 103: Please adjust the README too - that says 3.7+

Also, what is the reason for removing Python 3.7 support? Is it because Pandas 1.5 doesn't support it?

valeriy42: Yes, many new versions of the common package (pandas, numpy, matplotlib, scikit-learn) dropped support of Python 3.7. To keep it, we would basically require to keep wide support of package versions, which disproportionally increases the support overhead. Also, testing old packages on M1 hardware is technically impossible or at least very cumbersome.

docs/requirements-docs.txt
valeriy42: Fixed in [b07d145](https://github.com/elastic/eland/pull/512/commits/b07d14551ca1c8fc531563abfb1f12b3761d7f5c)

noxfile.py
valeriy42: README fixed in [b07d145](https://github.com/elastic/eland/pull/512/commits/b07d14551ca1c8fc531563abfb1f12b3761d7f5c)

sethmlarson (approved):
LGTM

Issue #514, opened 2023-02-06
Improve `task_type_from_model_config` function to auto determine model task types


jeffvestal:
Currently when downloading a model from Hugging Face with `TransformerModel`, `setting the task_type` to `auto` doesn't handle model configs where the architecture is not `None`

Managing different architecture types with `auto` will improve the use for end users rather than always having to specify a specific type. 

Example:
[msmarco-MiniLM-L-12-v3](https://huggingface.co/sentence-transformers/msmarco-MiniLM-L-12-v3) should be able to be detected as a `text_embedding` however the model architecture is listed as:
```
SentenceTransformer(
  (0): Transformer({'max_seq_length': 512, 'do_lower_case': False}) with Transformer model: BertModel 
  (1): Pooling({'word_embedding_dimension': 384, 'pooling_mode_cls_token': False, 'pooling_mode_mean_tokens': True, 'pooling_mode_max_tokens': False, 'pooling_mode_mean_sqrt_len_tokens': False})
)
```
so an error is raised by eland

```
TaskTypeError                             Traceback (most recent call last)
[<ipython-input-27-26a3b6b90e6f>](https://localhost:8080/#) in <module>
      1 model_id='distilbert-base-cased-distilled-squad'
      2 #tm = TransformerModel("sentence-transformers/msmarco-MiniLM-L-12-v3", "text_embedding")
----> 3 tm = TransformerModel("sentence-transformers/msmarco-MiniLM-L-12-v3", "auto")

1 frames
[/usr/local/lib/python3.8/dist-packages/eland/ml/pytorch/transformers.py](https://localhost:8080/#) in _create_traceable_model(self)
    629             maybe_task_type = task_type_from_model_config(model.config)
    630             if maybe_task_type is None:
--> 631                 raise TaskTypeError(
    632                     f"Unable to automatically determine task type for model {self._model_id}, please supply task type: {SUPPORTED_TASK_TYPES_NAMES}"
    633                 )

TaskTypeError: Unable to automatically determine task type for model sentence-transformers/msmarco-MiniLM-L-12-v3, please supply task type: fill_mask, ner, question_answering, text_classification, text_embedding, zero_shot_classification
```

PR #516, opened 2023-02-08
Choose text_embedding from auto when task type is unknown but its a sentence-transfomers model

benwtrent:
closes https://github.com/elastic/eland/issues/514


droberts195 (approved):
LGTM

PR #526, opened 2023-04-06
[NLP] Add support for the pass_through task

davidkyle:
The `pass_through` task returns the models output directly without any post-processing of the results. `pass_through` is typically used for debugging. 

This change allows models to be loaded into Elasticsearch with the `pass_through` task

An example invocation of the `eland_import_hub_model` script with `pass_through` is: 
```
eland_import_hub_model \
      --url XXX \
      -u XXX -p XXXX \
      --hub-model-id "sentence-transformers/msmarco-MiniLM-L-12-v3" \
      --task-type pass_through 
```


droberts195 (approved):
LGTM

PR #532, opened 2023-04-24
Set embedding size for Text Embedding models

davidkyle:
elastic/elasticsearch#95176 added the `embedding_size` field to text embedding model configs. The field is set at creation and cannot be changed.  

This change now sets `embedding_size` when creating the config for text embedding models. The embedding size is found by evaluating the model and taking the length of the last dimension of the output tensor


droberts195 (approved):
LGTM

PR #535, opened 2023-05-02
Check model output when measuring embedding size

### LINKED ISSUE #533: Getting "ValueError: not enough values to unpack" when using text_embedding models
Hello!

I'm getting the following error when trying to deploy a model with the task type "text_embedding"

image

Here is the command I used:

docker run -it --rm --network host \
    elastic/eland \
    eland_import_hub_model \
      --url "my_url" \
      -u "my_usrnm" -p "my_pswd" \
      --hub-model-id sentence-transformers/msmarco-MiniLM-L-12-v3 \
      --task-type text_embedding \
      --start

I tried all-mpnet-base-v2 as well as some other models and got the same error. Strangely task type text_classification works just fine.

I tried changing line 653 in eland/ml/pytorch/transformers.py so that it unpacks one value only but got a further api error.

Any suggestions are appreciated!

davidkyle:
#532 added code for generating a text_embedding and measuring the number of dimensions to set the `embedding_size` field. The code expected the model output to be a tuple but for some models that is not the case.

Closes #533

### DIFF
--- a/eland/ml/pytorch/transformers.py
+++ b/eland/ml/pytorch/transformers.py
@@ -650,8 +650,13 @@ class TransformerModel:
                 classification_labels=self._traceable_model.classification_labels(),
             )
         elif self._task_type == "text_embedding":
-            sample_embedding, _ = self._traceable_model.sample_output()
-            embedding_size = sample_embedding.size(-1)
+            sample_embedding = self._traceable_model.sample_output()
+            if type(sample_embedding) is tuple:
+                text_embedding, _ = sample_embedding
+            else:
+                text_embedding = sample_embedding
+
+            embedding_size = text_embedding.size(-1)
             inference_config = TASK_TYPE_TO_INFERENCE_CONFIG[self._task_type](
                 tokenization=tokenization_config,
                 embedding_size=embedding_size,


sethmlarson (commented):
No issues from my end, but is there a way we can add a test case for this?

eland/ml/pytorch/transformers.py
```diff
@@ -573,7 +573,13 @@ def _prepare_inputs(self) -> transformers.BatchEncoding:
 
 
 class TransformerModel:
-    def __init__(self, model_id: str, task_type: str, quantize: bool = False):
+    def __init__(
+        self,
+        model_id: str,
```
sethmlarson, line 578: Can we change this API to only accept keyword arguments? Then we can add new arguments whenever we want without worrying about positioning weirdness.

eland/ml/pytorch/transformers.py
```diff
@@ -573,7 +573,13 @@ def _prepare_inputs(self) -> transformers.BatchEncoding:
...
+        task_type: str,
+        es_version: Tuple[int, int, int],
```
sethmlarson, line 580: I'm a little apprehensive about forcing users to pass an `es_version` directly into this class, is there any other way we can handle this? (Should we wait to do any version-specific logic until the model needs to be serialized to Elasticsearch, at which point we'll have a client instance?)

sethmlarson (commented):
Thanks! Here are some thoughts:

eland/ml/pytorch/transformers.py
davidkyle: ++ I learned a new trick today

eland/ml/pytorch/transformers.py
davidkyle: The `TransformerModel` class is responsible for loading the model and converting it to a format that can be uploaded to Elasticsearch. The `PyTorchModel` class does the actual upload and has the client. Yes is would be nice to not require the version here but it would require a significant refactoring of those 2 classes that would probably break existing usages.

The version is required so that new features can be supported and to avoid creating incompatible configurations on older clusters, I expect we will see more of this type of logic in the future.

I've made the version parameter optional so that upgrade won't break any existing scripts using the class. If the version is not set only the most basic settings are used. I've also added docs explaining the usage. 

Another option is to pass the client rather than the cluster version to the `TransformerModel` class but then you need a working client and cluster to get the version when the use case might be just to save the model locally.

eland/ml/pytorch/transformers.py
```diff
@@ -651,12 +683,24 @@ def _create_config(self) -> NlpTrainedModelConfig:
...
         elif self._task_type == "text_embedding":
-            sample_embedding, _ = self._traceable_model.sample_output()
-            embedding_size = sample_embedding.size(-1)
-            inference_config = TASK_TYPE_TO_INFERENCE_CONFIG[self._task_type](
-                tokenization=tokenization_config,
-                embedding_size=embedding_size,
-            )
+            # The embedding_size paramater was added in Elasticsearch 8.8
+            # If the version is not known use the basic config
+            if es_version is None or (es_version[0] <= 8 and es_version[1] < 8):
```
sethmlarson, line 688: Let's simplify this to `es_version is None or es_version < (8, 8)`

eland/ml/pytorch/transformers.py
sethmlarson: Sounds good, if you're happy with the implementation then I am okay with it.

sethmlarson (approved):
Had one nit, otherwise LGTM!

PR #540, opened 2023-05-22
Limit NumPy to a range of versions and add note

davidkyle:
Follow up to #539 this adds a comment explaining why the NumPy versions has been limited and sets a lower bound


dolaru (approved):
LGTM 👍🏽

davidkyle:
Thanks for the review @dolaru

PR #541, opened 2023-05-22
Generate valid model id from file path

davidkyle:
The `eland_import_hub_model` script supports uploading a local file where the `--hub-model-id` argument is a file path. 

If the `--es-model-id` option is not used the model Id is generated from the hub model id and when that is a file path the path must be converted to a valid elasticsearch model id.

- Replace whitespace, forward slash and back slash with double underscore '__'
- Lower case
- Left truncated to 64 chars


valeriy42 (approved):
LGTM

PR #543, opened 2023-05-23
Add Elasticsearch 8.8 snapshot to test matrix

davidkyle:
One of the tests started failing in 8.8 with a Circuit Breaker Exception from Elasticsearch, likely due to better memory accounting in https://github.com/elastic/elasticsearch/issues/89437. The test node JVM heap is bumped to 2GB in response to the better memory accounting


valeriy42 (approved):
LGTM

PR #548, opened 2023-06-15
Fix autosummary directive by removing hack autosummaries

sethmlarson:
I'm unsure what these hack autosummaries are for because the documentation generates properly for the pages once they've been removed. They were added [4 years ago in the initial documentation commit](https://github.com/elastic/eland/commit/e181476dfe25a2abbe3dcf16ea663d2cbf09eab2#diff-b737ddab0ea189a06b340c34e8dca2352b1b3a55df783308abcf9094cadfba17R12) and haven't changed since. I think these are okay to remove as they're obviously incorrect content-wise.

### DIFF
--- a/docs/sphinx/conf.py
+++ b/docs/sphinx/conf.py
@@ -58,6 +58,7 @@ release = version
 # ones.
 extensions = [
     "sphinx.ext.autodoc",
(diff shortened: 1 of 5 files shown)


szabosteve (approved):
LGTM!

Issue #555, opened 2023-07-05
[CI] Model exporter test failure due to SK Learn format change 


davidkyle:
`test_ml_model_pytest` model fails because the tree format is not the expected type.

This probably due to a change in the recent [1.3 release](https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_3_0.html)

```
23:48:36 tests/ml/test_ml_model_pytest.py:615: 
23:48:36 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
23:48:36 eland/ml/ml_model.py:466: in export_model
23:48:36     model = ESGradientBoostingRegressor(
23:48:36 eland/ml/exporters/es_gb_models.py:393: in __init__
23:48:36     ESGradientBoostingModel.__init__(self, es_client, model_id)
23:48:36 eland/ml/exporters/es_gb_models.py:114: in __init__
23:48:36     self._trees.append(Tree(trained_model["tree"], feature_names_map))
23:48:36 eland/ml/exporters/_sklearn_deserializers.py:119: in __init__
23:48:36     self.tree.__setstate__(state)
23:48:36 sklearn/tree/_tree.pyx:714: in sklearn.tree._tree.Tree.__setstate__
23:48:36     ???
23:48:36 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
23:48:36 
23:48:36 >   ???
23:48:36 E   ValueError: node array from the pickle has an incompatible dtype:
23:48:36 E   - expected: {'names': ['left_child', 'right_child', 'feature', 'threshold', 'impurity', 'n_node_samples', 'weighted_n_node_samples', 'missing_go_to_left'], 'formats': ['<i8', '<i8', '<i8', '<f8', '<f8', '<i8', '<f8', 'u1'], 'offsets': [0, 8, 16, 24, 32, 40, 48, 56], 'itemsize': 64}
23:48:36 E   - got     : [('left_child', '<i8'), ('right_child', '<i8'), ('feature', '<i8'), ('threshold', '<f8'), ('impurity', '<f8'), ('n_node_samples', '<i8'), ('weighted_n_node_samples', '<f8')]
```

PR #558, opened 2023-07-10
Fix failing unit tests

valeriy42:
I updated the tree serialization format for the new scikit learn versions. I also updated the minimum requirement of scikit learn to 1.3 to ensure compatibility.

Fixes https://github.com/elastic/eland/issues/555

### DIFF
--- a/tests/ml/pytorch/test_pytorch_model_upload_pytest.py
+++ b/tests/ml/pytorch/test_pytorch_model_upload_pytest.py
@@ -105,6 +105,8 @@ def download_model_and_start_deployment(tmp_dir, quantize, model_id, task):
 class TestPytorchModel:
     def __init__(self):
         # quantization does not work on ARM processors
+        # TODO: It seems that PyTorch 2.0 supports OneDNN for aarch64. We should 
+        # revisit this when we upgrade to PyTorch 2.0.
         import platform
 
         self.quantize = (
--- a/tests/ml/pytorch/test_pytorch_model_upload_pytest.py
+++ b/tests/ml/pytorch/test_pytorch_model_upload_pytest.py
@@ -103,23 +103,31 @@ def download_model_and_start_deployment(tmp_dir, quantize, model_id, task):
 
 
 class TestPytorchModel:
-
     def __init__(self):
         # quantization does not work on Mac with ARM processor
         import platform
-        self.quantize = True if not (platform.system() == "Darwin" and platform.machine() == "arm64") else False
+
+        self.quantize = (
+            True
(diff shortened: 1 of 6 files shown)


tests/ml/pytorch/test_pytorch_model_upload_pytest.py
```diff
@@ -103,17 +103,31 @@ def download_model_and_start_deployment(tmp_dir, quantize, model_id, task):
 
 
 class TestPytorchModel:
+    def __init__(self):
+        # quantization does not work on Mac with ARM processor
+        import platform
+
+        self.quantize = (
+            True
+            if not (platform.system() == "Darwin" and platform.machine() == "arm64")
```
droberts195, line 112: Is it just ARM on Mac? Or would ARM on Linux be the same?

(If you don't know, you could hire a `c6g` instance in AWS to find out.)

droberts195 (approved):
LGTM, although I had a question about whether it's just ARM Mac that cannot do quantization.

tests/ml/pytorch/test_pytorch_model_upload_pytest.py
valeriy42: You are right, I had a Linux docker container for arm64 platform and it didn't support oneDNN either. Things seem to get better in Pytorch 2.0 so.

tests/ml/pytorch/test_pytorch_model_upload_pytest.py
```diff
@@ -103,17 +103,27 @@ def download_model_and_start_deployment(tmp_dir, quantize, model_id, task):
 
 
 class TestPytorchModel:
+    def __init__(self):
+        # quantization does not work on ARM processors
+        import platform
+
+        self.quantize = True if platform.machine() != "arm64" else False
```
droberts195, line 110: Does `platform.machine()` return `aarch64` on ARM Linux? (It will if it simply returns the output of `uname -m`.) If so, please change the condition to check for `platform.machine()` being either `arm64` or `aarch64`.

tests/ml/pytorch/test_pytorch_model_upload_pytest.py
```diff
@@ -103,17 +103,27 @@ def download_model_and_start_deployment(tmp_dir, quantize, model_id, task):
 
 
 class TestPytorchModel:
+    def __init__(self):
+        # quantization does not work on ARM processors
```
droberts195, line 107: If you think the situation is better in PyTorch 2.0 please add a TODO to revisit this when we upgrade.

droberts195 (approved):
LGTM - thanks for making the edits

PR #560, opened 2023-07-18
[Backport][7.x][CI] Migrates CI to Buildkite

picandocodigo:

PR #559, opened 2023-07-13
Migrate to entry_points #551

sakurai-youhei:
Closes #551.

With this PR, `eland_import_hub_model` gets easier to find on Windows.

**On windows**: `eland_import_hub_model` is an executable file.

```console
C:\Users\YouheiSakurai\git\myeland>where eland_import_hub_model
C:\Users\YouheiSakurai\AppData\Roaming\Python\Python311\Scripts\eland_import_hub_model.exe
```

**On Linux**: `eland_import_hub_model` is a simple auto-generated script file.

```console
root@b16946fa4fe1:/# which eland_import_hub_model
/usr/local/bin/eland_import_hub_model

root@b16946fa4fe1:/# file `which eland_import_hub_model`
/usr/local/bin/eland_import_hub_model: Python script, ASCII text executable

root@b16946fa4fe1:/# head `which eland_import_hub_model`
#!/usr/local/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from eland.cli.eland_import_hub_model import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

Note: Because I come to think adding `python -m eland` would be too ambicious, I dropped it from the scope.

PR #562, opened 2023-07-18
[7.x] Format code for linting failures

davidkyle:
Also pins the Elasticsearch client <8 for the 7.x branch

Follow up to #560

### DIFF
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -1,4 +1,4 @@
-elasticsearch>=7.7
+elasticsearch>=7.7,<8
 pandas>=1.2.0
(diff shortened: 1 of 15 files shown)


picandocodigo:
@davidkyle I've just enabled Buildkite builds for pull requests from forks since this wasn't building. But I think the PR might need to be updated for Buildkite to pick up the change and trigger the build.

davidkyle:
> I've just enabled Buildkite builds for pull requests from forks 

Ah I guess that is why the CI on https://github.com/elastic/eland/pull/559 hung. Thanks @picandocodigo

PR #567, opened 2023-07-20
[ML] Better memory estimation for NLP models

valeriy42:


valeriy42:
PR is cleaned up in #568

PR #571, opened 2023-07-26
Simplify embedding model support and loading

### LINKED ISSUE #531: [NLP] Add option for using the pooling layer in Text Embedding models
BERT based text embedding models usually output a pooling layer and the last hidden state. The first element of the last hidden state is the CLS token which is the embedding representation the entire input. Conventionally the last hidden state is the first output of the model and this is what is returned by the C++ code.

https://github.com/elastic/ml-cpp/blob/main/bin/pytorch_inference/Main.cc#L63

The Java then takes the first vector from the tensor which is the CLS token

https://github.com/elastic/elasticsearch/blob/a8d57233093385169d158e931995a50bf9a135aa/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/TextEmbeddingProcessor.java#LL57C54-L57C54

It takes quite a deep dive to figure out that ES is returning the CLS token and not the output of some pooling layer. There is some debate as to which is best to use, some might prefer to use a pooling layer and others the CLS token.

To avoid painful code changes in C++ and Java it would be easiest to wrap the model to select preferred output in Eland but making it the first item returned.

Note: Eland removes the pooling layer from Sentence Transformer models 
https://github.com/elastic/eland/blob/main/eland/ml/pytorch/transformers.py#L306

joshdevins:
We were attempting to load SentenceTransformers by looking at the model prefix, however SentenceTransformers can also be loaded from other orgs in the model hub, as well as from local disk. This prefix checking failed in those two cases. To simplify the loading logic and deciding which wrapper to use, we’ve removed support for text_embedding tasks to load a plain Transformer. We now only support DPR embedding models and SentenceTransformer embedding models. If you try to load a plain Transformer model, it will be loaded by SentenceTransformers and a mean pooling layer will automatically be added by the SentenceTransformer library. Since we no longer automatically support non-DPR and non-SentenceTransformers, we should include somewhere example code for how to load a custom model without DPR or SentenceTransformers. 

Note: This change will allow us to support E5 embeddings uploaded with eland. This change will _not_ yet solve adding the preamble instructions to query and index embeddings.

See: https://github.com/UKPLab/sentence-transformers/blob/v2.2.2/sentence_transformers/SentenceTransformer.py#L801

Resolves #531

### DIFF
--- a/eland/ml/pytorch/transformers.py
+++ b/eland/ml/pytorch/transformers.py
@@ -294,22 +294,19 @@ class _SentenceTransformerWrapperModule(nn.Module):  # type: ignore
     def from_pretrained(
         model_id: str, output_key: str = DEFAULT_OUTPUT_KEY
     ) -> Optional[Any]:
-        if model_id.startswith("sentence-transformers/"):
-            model = AutoModel.from_pretrained(model_id, torchscript=True)
-            if isinstance(
-                model.config,
-                (
-                    transformers.MPNetConfig,
-                    transformers.XLMRobertaConfig,
-                    transformers.RobertaConfig,
-                    transformers.BartConfig,
-                ),
-            ):
-                return _TwoParameterSentenceTransformerWrapper(model, output_key)
-            else:
-                return _SentenceTransformerWrapper(model, output_key)
+        model = AutoModel.from_pretrained(model_id, torchscript=True)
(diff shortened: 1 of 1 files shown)


joshdevins:
Ignore this PR

PR #573, opened 2023-08-01
Error measuring embedding size for some DPR models

davidkyle:
Installing the `ance-dpr-context-multi` model with the `eland_import_hub_model` script fails. 

``` 
eland_import_hub_model       
    --url 'http://localhost:9200'             
    --hub-model-id 'ance-dpr-context-multi'
    --task-type text_embedding 
```

The error is due to unpacking a tuple that contains a single element. 
```
python3.10/site-packages/eland/ml/pytorch/transformers.py", line 735, in _create_config
    text_embedding, _ = sample_embedding
ValueError: not enough values to unpack (expected 2, got 1)
```

The fix is to only take the first element


eland/ml/pytorch/transformers.py
```diff
@@ -732,7 +732,7 @@ def _create_config(
             else:
                 sample_embedding = self._traceable_model.sample_output()
                 if type(sample_embedding) is tuple:
-                    text_embedding, _ = sample_embedding
+                    text_embedding, *_ = sample_embedding
```
joshdevins, line 735: This works but it's maybe obscure Python code. We don't actually need to unpack the whole tuple, so it might make it clearer to just extract the first element, which is what we actually care about:

```python
text_embedding = sample_emebdding[0]
```

joshdevins (changes requested):

joshdevins (approved):

davidkyle:
buildkite test this

Issue #580, opened 2023-08-28
Action monitor/main is unauthorized for API key


bartbroere:
When using `eland` with an Elasticsearch instance that authorizes using an API key, I often get this error:

```
elasticsearch.AuthorizationException: AuthorizationException(403, "security_exception", "action [cluster:monitor/main] is unauthorized for API key id [...] of user [...], this action is granted by the cluster privileges [monitor,manage,all]",)
```

This is triggered because some functionality is switched on or off depending on the version of the Elasticsearch you are connecting to.

I can solve it in my code by first making the Elasticsearch connection, and then modifying its attribute `_eland_es_version`, and set it to the version number I'm using.

```python
import elasticsearch
from eland import DataFrame

elastic_client = elasticsearch.Elasticsearch(
    "https://omitted.us-central1.gcp.cloud.es.io:443",
    api_key="...")

elastic_client._eland_es_version = (8, 9, 1)

df = DataFrame(elastic_client,
               'search-*')
```

Is there a different way to check the version of the Elasticsearch host, that requires less permissions perhaps?

Here's the code that determines the version number currently: https://github.com/elastic/eland/blob/f14bbaf4b0ed072ce2c74cfb2511c25c3c547cd6/eland/common.py#L318C1-L341

droberts195:
Action `cluster:monitor/main` corresponds to the root REST endpoint of Elasticsearch, i.e. `/`.

The clients use this to determine whether they're talking to Elasticsearch.

I think the best solution is to give your API keys permission to call this endpoint. Instead of granting a high level cluster privilege that would give wide access you should be able to just add `cluster:monitor/main` to the cluster privileges you're currently granting. (It's possible to use individual action names in addition to high level privilege names when creating role definitions.)

bartbroere:
@droberts195 Thanks for your reply. For an API key, setting a wider scope might be a solution. The first time I encountered this issue however, was in a corporate environment, with an Elasticsearch deployment linked to LDAP. In that case, the user might not have the luxury to be able to grant additional permissions.

In #581 I propose a small change that falls back to a default code path if the server's version cannot be determined, instead of raising the error. Maybe this can be a nice solution.

PR #581, opened 2023-08-30
Use the workaround if we can't determine the server's version

bartbroere:
This is related to issue #580 

In some scenarios (a limited-scope API key, or maybe a corporate environment) a user of `eland` might not have enough permissions to determine the host's Elasticsearch version.

In that case, we could default to using the workaround, to be sure.

### DIFF
--- a/eland/field_mappings.py
+++ b/eland/field_mappings.py
@@ -948,10 +948,12 @@ def _compat_field_caps(client, fields, index=None):
     # If we lack sufficient permission to determine the Elasticsearch version,
     # to be sure we use the workaround for versions smaller than 8.5.0
     except elasticsearch.AuthorizationException as e:
-        raise RuntimeWarning("Couldn't determine Elasticsearch host's version. "
-                             "Probably missing monitor/main permissions. "
-                             "Continuing with the query string work-around. "
-                             "Original exception: " + str(e))
+        raise RuntimeWarning(
+            "Couldn't determine Elasticsearch host's version. "
+            "Probably missing monitor/main permissions. "
+            "Continuing with the query string work-around. "
+            "Original exception: " + str(e)
+        )
         elastic_version = None
     if elastic_version and elastic_version >= (8, 5, 0):
         return client.field_caps(index=index, fields=fields)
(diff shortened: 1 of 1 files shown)


pquentin:
buildkite test this please

pquentin:
buildkite test this please

eland/field_mappings.py
```diff
@@ -942,7 +943,19 @@ def _compat_field_caps(client, fields, index=None):
...
+    try:
+        elastic_version = es_version(client)
+    # If we lack sufficient permission to determine the Elasticsearch version,
+    # to be sure we use the workaround for versions smaller than 8.5.0
+    except elasticsearch.AuthorizationException as e:
+        raise RuntimeWarning(
+            "Couldn't determine Elasticsearch host's version. "
+            "Probably missing monitor/main permissions. "
+            "Continuing with the query string work-around. "
+            "Original exception: " + str(e)
```
pquentin, line 955:
```suggestion
            "Original exception: " + repr(e)
```

pquentin (approved):
I'm happy to do this is it unblocks you and the tests pass, but note that there are other places where we look at the ES version in the same way that would break similarly.

bartbroere:
> I'm happy to do this is it unblocks you, but note that there are other places where we look at the ES version in the same way that would break similarly.

Thanks! I noticed too that it was used in some other places. This one is probably in the place someone is most likely to encounter it. You get to this code by instantiating an eland DataFrame I believe.

Of course I'm still trying to find out if we could grant `monitor/main` privileges to all users in our organisation, but for now this would be very helpful.

eland/field_mappings.py
bartbroere: Thanks, this is the better way indeed!

droberts195:
> Of course I'm still trying to find out if we could grant `monitor/main` privileges to all users in our organisation

`cluster:monitor/main` literally just gives access to the `/` endpoint, which I find hard to believe is a security risk to allow for any user who legitimately has access to the cluster to do anything else. We probably ought to make it easier to grant access to this endpoint alone instead of requiring an obscure action name to be granted. But until then I don’t think it would be a security risk to add it manually to some of the other roles you have defined.

PR #582, opened 2023-08-31
Update README.md with note about installing extras for NLP

davidkyle:


pquentin (approved):
Thanks for clarifying the README! LGTM.

PR #584, opened 2023-09-04
Fix README so that copy/pastes work without warnings

pquentin:
* Using `basic_auth` prints the following warning since elasticsearch-py 8.0:

    ```python
    DeprecationWarning: The 'http_auth' parameter is deprecated. Use 'basic_auth' or 'bearer_auth' parameters instead
    ```
* Using "localhost:9200" fails outright since elasticsearch-py 8.0. I wasn't sure if `http` or `https` should be used, but the rest of the repo uses `http`.

    ```python
    ValueError: URL must include a 'scheme', 'host', and 'port' component (ie 'https://localhost:9200')
    ```
* The XGBoost section assumes `training_data` exists: let's help users by creating it for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,10 +205,12 @@ libraries to be serialized and used as an inference model in Elasticsearch.
 ➤ [Read more about Machine Learning in Elasticsearch](https://www.elastic.co/guide/en/machine-learning/current/ml-getting-started.html)
 
 ```python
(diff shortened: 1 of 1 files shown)


davidkyle (approved):
LGTM 

Thanks @pquentin

PR #587, opened 2023-09-04
Fix downloads badge URL

pquentin:
You can see the result here: https://github.com/pquentin/eland/tree/badge-url.


davidkyle (approved):
LGTM

Now it's fixed I see 173k downloads! pretty good!

PR #590, opened 2023-09-05
Add dummy pipeline to prepare publishing a Docker image

pquentin:
The idea is to get the scaffolding in place before testing the pipeline itself. This also declares the existing pipeline in the repository itself, according to latest Buildkite good practices at Elastic.


JoshMock (approved):
looks good!

PR #596, opened 2023-09-11
Remove deprecated check_less_precise

bartbroere:
This removes the deprecated `check_less_precise` argument from one of the unit tests.

Part of the necessary adjustments to maybe start supporting `pandas>=2.0.0`

### DIFF
--- a/tests/series/test_arithmetics_pytest.py
+++ b/tests/series/test_arithmetics_pytest.py
@@ -74,7 +74,8 @@ class TestSeriesArithmetics(TestData):
             ed_series.__class__ = ModifiedElandSeries
 
             assert_pandas_eland_series_equal(
-                pd_series, ed_series, check_less_precise=True
+                pd_series, ed_series,
+                rtol=1e-3  # previously known as check_less_precise
             )
 
     def test_ecommerce_series_invalid_div(self):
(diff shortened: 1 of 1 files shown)


pquentin (changes requested):
Do we also need to specify `check_exact=False`? According to the 2.0 docs, [rtol is defined as](https://pandas.pydata.org/pandas-docs/version/2.0/reference/api/pandas.testing.assert_series_equal.html):

> Relative tolerance. Only used when check_exact is False.

bartbroere:
> Do we also need to specify `check_exact=False`? According to the 2.0 docs, [rtol is defined as](https://pandas.pydata.org/pandas-docs/version/2.0/reference/api/pandas.testing.assert_series_equal.html):
> 
> > Relative tolerance. Only used when check_exact is False.

Good catch! I'll update it.

bartbroere:
> Do we also need to specify `check_exact=False`? According to the 2.0 docs, [rtol is defined as](https://pandas.pydata.org/pandas-docs/version/2.0/reference/api/pandas.testing.assert_series_equal.html):
> 
> > Relative tolerance. Only used when check_exact is False.



> > Do we also need to specify `check_exact=False`? According to the 2.0 docs, [rtol is defined as](https://pandas.pydata.org/pandas-docs/version/2.0/reference/api/pandas.testing.assert_series_equal.html):
> > > Relative tolerance. Only used when check_exact is False.
> 
> Good catch! I'll update it.

@pquentin I added the `check_exact=False` flag, but Github still doesn't think I "addressed the requested changes". Not sure how that works,  but I think it's all good now.

pquentin:
buildkite test this please

PR #597, opened 2023-09-11
Add pandas 1.4.4 to test matrix

### LINKED ISSUE #451: Ensure support for pandas 1.4.x
May be a no-op, but we need to run our test suite with pandas 1.4.x to make sure.

bartbroere:
Closes #451

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -101,7 +101,10 @@ def lint(session):
 
 
 @nox.session(python=["3.8", "3.9", "3.10"])
-@nox.parametrize("pandas_version", ["1.5.0"])
+@nox.parametrize("pandas_version", [
+    "1.4.4",
+    "1.5.0",
+])
 def test(session, pandas_version: str):
     session.install("-r", "requirements-dev.txt")
     session.install(".")


pquentin (commented):
Hello, and thanks for the many contributions!

This change itself won't have any effect, because Eland requires Pandas 1.5+ since #512. We could add back support, but Pandas 1.4 isn't particularly popular compared to Pandas 2.x or even Pandas 1.3. Screenshot of https://www.pepy.tech/projects/pandas?versions=2.*&versions=1.5.*&versions=1.4.*&versions=1.3.* as of today:

image

@valeriy42 @davidkyle I think we should focus our efforts on Pandas 2.x instead, what do you think?

bartbroere:
Feel free to close this PR in that case. 

My other pull requests are indeed contributions to start being able to support Pandas 2.0. For me that also is more valuable than 1.4 support. I want to be able to introduce eland in projects at work, without having to pin pandas to an older version in these projects.

davidkyle:
I agree let's close this PR and merge the changes to support Pandas 2.0

PR #598, opened 2023-09-11
Start supporting Python 3.11

bartbroere:
The issue of PyTorch not supporting Python 3.11 (pytorch/pytorch#86566) seems to have been resolved. I did not see any apparent other issues with supporting Python 3.11 in the code.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -16,13 +16,11 @@ scikit-learn>=1.3,<2
 xgboost>=0.90,<2
 lightgbm>=2,<4
 
-# PyTorch doesn't support Python 3.11 yet (pytorch/pytorch#86566)
-
 # Elasticsearch uses v1.13.1 of PyTorch
-torch>=1.13.1,<2.0; python_version<'3.11'
+torch>=1.13.1,<2.0
 # Versions known to be compatible with PyTorch 1.13.1
-sentence-transformers>=2.1.0,<=2.2.2; python_version<'3.11'
-transformers[torch]>=4.12.0,<=4.27.4; python_version<'3.11'
+sentence-transformers>=2.1.0,<=2.2.2
+transformers[torch]>=4.12.0,<=4.27.4
 
 #
 # Testing
--- a/noxfile.py
+++ b/noxfile.py
@@ -100,7 +100,7 @@ def lint(session):
             session.error("\n" + "\n".join(sorted(set(errors))))
 
 
-@nox.session(python=["3.8", "3.9", "3.10"])
+@nox.session(python=["3.8", "3.9", "3.10", "3.11"])
 @nox.parametrize("pandas_version", ["1.5.0"])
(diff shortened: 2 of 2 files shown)


requirements-dev.txt
```diff
@@ -16,13 +16,11 @@ scikit-learn>=1.3,<2
 xgboost>=0.90,<2
 lightgbm>=2,<4
 
-# PyTorch doesn't support Python 3.11 yet (pytorch/pytorch#86566)
-
 # Elasticsearch uses v1.13.1 of PyTorch
-torch>=1.13.1,<2.0; python_version<'3.11'
+torch>=1.13.1,<2.0
```
davidkyle, line 20: I think it is worth keeping the restriction for the unreleased Python '3.12' as we do not know if that will be compatible with PyTorch yet. 

```suggestion
torch>=1.13.1,<2.0; python_version<'3.12'
```

davidkyle (commented):
Thanks for the change @bartbroere. Looks good but I'd like to test this on a couple of platforms first

requirements-dev.txt
bartbroere: > I think it is worth keeping the restriction for the unreleased Python '3.12' as we do not know if that will be compatible with PyTorch yet.

Good point, I applied the suggestion.

davidkyle:
Testing this I found it is not possible to install torch 1.13.1 with Python3.11.

Using pip install with Python 3.10 the install is successful:
```
/Library/Frameworks/Python.framework/Versions/3.10/bin/python3.10 -m pip install torch==1.13.1
```

With Python 3.11 the same install fails
```
/Library/Frameworks/Python.framework/Versions/3.11/bin/python3.11 -m pip install torch==1.13.1
```
```
ERROR: Could not find a version that satisfies the requirement torch==1.13.1 (from versions: 2.0.0, 2.0.1)
ERROR: No matching distribution found for torch==1.13.1
```

It looks like PyTorch 1.13.1 does not support Python 3.11, we will have to upgrade to PyTorch 2.0 for that support

@bartbroere did you get a different result?

pquentin:
PyTorch 1.13.1 only supports Python 3.11 on Linux, not on Windows or macOS, per the "cp311" files in https://pypi.org/project/torch/1.13.1/#files. It seems OS to support Python 3.11 on Linux but not on other platforms, but I'm biased as a Linux user :)

davidkyle:
Thanks @pquentin, allowing the option of using Python 3.11 generally will break model upload for too many users. 

@bartbroere unless you have an objection I'd like to close this PR without merging as neither macOS or Windows users can use Python 3.11 with PyTorch 1.13.1

bartbroere:
Good discussion, and apologies I may not have had the right environment during testing this locally! I closed this pull request, since I agree it's wise not to merge this one until a newer Pytorch is supported.

PR #604, opened 2023-09-18
Allow running demo_notebook in Google Colab

pquentin:
Adding an "Open in Colab" button was easy, but making the actual notebook runnable in Colab took way more effort. Ive followed the conventions used in https://github.com/elastic/elasticsearch-labs but:

 * I've used a hack to allow loading the test data from that notebook.
 * Making this nicer (including using the cloud id instead of host/port) will require changes in the tests.

I was not sure who to ask for review, feel free to change the reviewers list.

### DIFF
--- a/docs/sphinx/examples/demo_notebook.ipynb
+++ b/docs/sphinx/examples/demo_notebook.ipynb
@@ -67,37 +67,47 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Collecting git+https://github.com/pquentin/eland.git@demo-notebook-colab\r\n",
-      "  Cloning https://github.com/pquentin/eland.git (to revision demo-notebook-colab) to /tmp/pip-req-build-llaufivk\r\n",
-      "  Running command git clone --filter=blob:none --quiet https://github.com/pquentin/eland.git /tmp/pip-req-build-llaufivk\r\n",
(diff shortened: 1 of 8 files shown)


pquentin:
Deleted Jenkins webhooks manually again, closing/reopening.

joshdevins (changes requested):
+1 to making it run in Colab.

I don't love cloning the repo to get access to tests. I think it might be cleaner to move the functionality we need out of the test and into some "data loader" module. This way we just install eland and the module is there. To get access to the data, we could just download it from the GitHub blob url.

pquentin:
@joshdevins What are your thoughts on this new approach? I've tested that `nox -s docs` works when ES is active, and that the notebook can be fully run in Colab too. Fetching the data from GitHub does seem like a nice approach, we're talking about less than 2MB here.

pquentin:
I've also enabled the Read the Docs PR builds, so you can see:

 * the new API: https://eland--604.org.readthedocs.build/en/604/reference/io.html
 * the full notebook: https://eland--604.org.readthedocs.build/en/604/examples/demo_notebook.html

joshdevins (approved):

pquentin:
Not going to merge this as is as I realized that `eland.etl` exposes various functions doing the same thing as mine.

pquentin:
Closing in favor of #630.

Issue #600, opened 2023-09-12
Stop installing GPU dependencies


pquentin:
Installing all Eland dependencies locally or in the Docker image installs 4.2GB of dependencies, the two largest being PyTorch (1.9GB) and the CUDA dependencies (https://pypi.org/project/nvidia-cudnn-cu11/, https://pypi.org/project/nvidia-cublas-cu11/, https://pypi.org/project/nvidia-cuda-nvrtc-cu11/ and https://pypi.org/project/nvidia-cuda-runtime-cu11/, taking 1.4GB).

Since we don't support GPUs, installing those is wasteful, and we should install the PyTorch variant that only supports CPUs:

```
pip install torch==1.13.1+cpu --extra-index-url https://download.pytorch.org/whl/cpu
```

It's not possible to specify the extra index URL when doing `python -m pip install eland[pytorch]`, but we could instead ask to run the above command. In any case, it's an easy fix in the Docker image which is probably the main way that PyTorhc is used with Eland today (and #407 will make this even more true).

PR #613, opened 2023-10-03
Add Buildkite pipeline to push Docker image

### LINKED ISSUE #407: Push eland Docker images with every version release
With the recently added Dockerfile (#405) for building a Docker image from the current/local eland, we introduce the ability to run eland installed scripts without installing eland or any prerequisites. This still requires building a local Docker image though. It would be nice to avoid this and simply make available a public Docker image from the Elastic Docker repository. This would make running scripts like uploading HuggingFace transfomer models a very simple one command, with no installation or builds necessary.

This would mean hooking the Docker build into the release process, and being able to push the image to the Elastic public Docker repository.

pquentin:
Closes #407

Here's a successful build: https://buildkite.com/elastic/eland-release-docker/builds/9#018af52e-7ca0-4e9c-8aa5-6fd9cd4348de. This also reintroduces https://github.com/elastic/eland/pull/586, using a slim Debian 12 base image, which will greatly reduce the noise from vulnerability scanners.

I've left #600 as future work.

### DIFF
--- a/.buildkite/release-docker/pipeline.yml
+++ b/.buildkite/release-docker/pipeline.yml
@@ -1,14 +1,28 @@
 steps:
   - input: "Build parameters"
     fields:
-      - text: "RELEASE_VERSION"
+      - text: "Release version"
         key: "RELEASE_VERSION"
         default: ""
-        hint: "The version to release e.g. '2.8.0'."
-
+        format: "\\d{1,}.\\d{1,}.\\d{1,}"
+        hint: "The version to release e.g. '8.10.0' (without the v prefix)."
+      - select: "Environment"
+        key: "ENVIRONMENT"
+        options:
+          - label: "Staging"
+            value: "staging"
+          - label: "Production"
+            value: "production"
   - wait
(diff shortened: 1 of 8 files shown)


ezimuel (approved):
LGTM!

pquentin:
So the lint runs with the latest Python, which as of yesterday is Python 3.12. While we could use Python 3.10 instead (the latest version we support), the lint errors do make sense and the fixes work in Python 3.8+. For the record:

 * Python 3.12 has improved f-string support, which apparently allowed flake8 to detect indentation errors in f-strings: [`e968cd6` (#613)](https://github.com/elastic/eland/pull/613/commits/e968cd66418f4c3b2ee1d75a25b78e0c3e0163ea)
 * Python 3.12 removed the distutils module that we were using to compare versions, this should be done with the third-party packaging model now: [`1dbe81d` (#613)](https://github.com/elastic/eland/pull/613/commits/1dbe81d4c0165fe03fd25e2b4ecc4a94948c48d7)

PR #615, opened 2023-10-03
Reduce Docker image size from 4.8GB to 2.2GB

### LINKED ISSUE #600: Stop installing GPU dependencies

pquentin:
Closes #600 

This was done by using the torchcpu PyPI package and not storing the PyPI cache in the Docker image.

Tested using `docker run -it --rm --network host -e ES_USERNAME=$ES_USERNAME -e ES_PASSWORD=$ES_PASSWORD eland eland_import_hub_model --cloud-id $CLOUD_ID  --hub-model-id elastic/distilbert-base-cased-finetuned-conll03-english --task-type ner`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@ RUN apt-get update && \
 ADD . /eland
 WORKDIR /eland
 
-RUN python3 -m pip install --no-cache-dir --disable-pip-version-check .[all]
+RUN --mount=type=cache,target=/root/.cache python3 -m pip install torch==1.13.1+cpu --extra-index-url https://download.pytorch.org/whl/cpu
+RUN --mount=type=cache,target=/root/.cache python3 -m pip install --no-cache-dir --disable-pip-version-check .[all]
 
 CMD ["/bin/sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM python:3.10-slim-bookworm@sha256:d364435d339ad318ac4c533b9fbe709739f9ba006b0721fd25e8592d0bb857cb
+# syntax=docker/dockerfile:1
+FROM python:3.10-slim
 
 RUN apt-get update && \
     apt-get install -y build-essential pkg-config cmake \
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,16 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 ADD . /eland
 WORKDIR /eland
 
-RUN --mount=type=cache,target=/root/.cache python3 -m pip install torch==1.13.1+cpu --extra-index-url https://download.pytorch.org/whl/cpu
-RUN --mount=type=cache,target=/root/.cache python3 -m pip install --no-cache-dir --disable-pip-version-check .[all]
+ARG TARGETPLATFORM
+RUN --mount=type=cache,target=/root/.cache/pip \
+    if [ "$TARGETPLATFORM" = "linux/amd64" ]; then \
+      python3 -m pip install \
+        --no-cache-dir --disable-pip-version-check --extra-index-url https://download.pytorch.org/whl/cpu  \
+        torch==1.13.1+cpu .[all]; \
+    else \
+      python3 -m pip install \
+        --no-cache-dir --disable-pip-version-check \
+        .[all]; \
+    fi
 
 CMD ["/bin/sh"]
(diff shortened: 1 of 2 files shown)


dolaru (commented):
@pquentin It looks like the `torch+cpu` variant is only available for Intel/AMD CPUs: 
```
torch-1.13.1+cpu-cp310-cp310-linux_x86_64.whl
torch-1.13.1+cpu-cp310-cp310-win_amd64.whl
```

This means the image won't build on systems running Docker Engine on Apple Silicon or other system with an ARM CPU. Are we comfortable with that restriction? 🤔 

_Later edit_: When building for `x86_64/amd64`, it makes sense to use the `torch+cpu` variant as it doesn't install the large NVIDIA deps we don't need. For `aarch64` though, with the normal `torch` package, the image size is only `1.3GB` so we don't need to rely on the `torch+cpu` variant.

Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM python:3.10-slim-bookworm@sha256:d364435d339ad318ac4c533b9fbe709739f9ba006b0721fd25e8592d0bb857cb
+# syntax=docker/dockerfile:1
+FROM python:3.10-slim
```
pquentin, line 2: What was the reason for using a loose tag? I'm starting to think it makes sense as it will allow to get the latest improvements for each release as long as we test the resulting Docker image to avoid surprises.

But still wondering if there was something else.

Dockerfile
```diff
@@ -13,7 +13,16 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 ADD . /eland
 WORKDIR /eland
 
-RUN --mount=type=cache,target=/root/.cache python3 -m pip install torch==1.13.1+cpu --extra-index-url https://download.pytorch.org/whl/cpu
-RUN --mount=type=cache,target=/root/.cache python3 -m pip install --no-cache-dir --disable-pip-version-check .[all]
+ARG TARGETPLATFORM
+RUN --mount=type=cache,target=/root/.cache/pip \
+    if [ "$TARGETPLATFORM" = "linux/amd64" ]; then \
+      python3 -m pip install \
+        --no-cache-dir --disable-pip-version-check --extra-index-url https://download.pytorch.org/whl/cpu  \
```
pquentin, line 20: `--no-cache-dir` is a good idea, thanks. That means we should remove `--mount=type=cache,target=/root/.cache/pip` above.

pquentin (commented):
Thanks! This is great.

The part that is left is building those images in Buildkite, modifying https://github.com/elastic/eland/blob/main/.buildkite/release-docker/run.sh. I've started reading https://docs.docker.com/build/building/multi-platform/ but it seems quite complicated. Do you have any suggestions about the best approach here?

Dockerfile
dolaru: I've left the cache mount in because `pip` still leaves something behind in `/root/.cache/pip` even with the `--no-cache-dir` option. 😞

pquentin:
Thanks to @dolaru, I was able to adapt run.sh to build and push an ARM image. This command worked:

```
ENVIRONMENT=staging RELEASE_VERSION=8.9.0 with-vault-ci bash .buildkite/release-docker/run.sh
```

I would appreciate if an Elastic employee with an Apple Silicon Macbook could test the resulting image. (Ping me offline to get instructions.)

dolaru:
Good stuff @pquentin! I can confirm that the ARM container image works as intended. 👍🏽 

I successfully imported the ` elastic/distilbert-base-cased-finetuned-conll03-english` model using `eland_import_hub_model` in that container image.

dolaru (approved):

PR #619, opened 2023-10-11
Fix direct usage of TransformerModel

### LINKED ISSUE #617: Make access_token an optional parameter in TransformerModel
#608 changed the `TransformerModel` constructor, making the `access_token` attribute required. Since tokens are not required for public models, users must now set `access_token` to an empty string to use the model. I think this is an unnecessary breaking change.

pquentin:
Closes #617

### DIFF
--- a/eland/ml/pytorch/transformers.py
+++ b/eland/ml/pytorch/transformers.py
@@ -584,7 +584,7 @@ class TransformerModel:
         self,
         *,
         model_id: str,
-        access_token: Optional[str],
+        access_token: Optional[str] = None,
         task_type: str,
         es_version: Optional[Tuple[int, int, int]] = None,
         quantize: bool = False,
--- a/eland/ml/pytorch/transformers.py
+++ b/eland/ml/pytorch/transformers.py
@@ -584,10 +584,10 @@ class TransformerModel:
         self,
         *,
         model_id: str,
-        access_token: Optional[str] = None,
         task_type: str,
         es_version: Optional[Tuple[int, int, int]] = None,
         quantize: bool = False,
+        access_token: Optional[str] = None,
     ):
         """
         Loads a model from the Hugging Face repository or local file and creates
--- a/noxfile.py
+++ b/noxfile.py
@@ -112,6 +112,8 @@ def test(session, pandas_version: str):
         "python",
         "-m",
         "pytest",
+        "-ra",
+        "--tb=native",
         "--cov-report=term-missing",
         "--cov=eland/",
         "--cov-config=setup.cfg",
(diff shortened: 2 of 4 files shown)


noxfile.py
```diff
@@ -112,6 +112,8 @@ def test(session, pandas_version: str):
         "python",
         "-m",
         "pytest",
+        "-ra",
```
pquentin, line 115: This reports all test except passes, which is useful to see why a specific would have been skipped.

noxfile.py
```diff
@@ -112,6 +112,8 @@ def test(session, pandas_version: str):
...
+        "--tb=native",
```
pquentin, line 116: This shows the Python native traceback, instead of the overly verbose pytest tracebacks that display the source code (I don't need it, I can see it in my editor).

requirements-dev.txt
```diff
@@ -22,7 +22,7 @@ lightgbm>=2,<4
 torch>=1.13.1,<2.0; python_version<'3.11'
 # Versions known to be compatible with PyTorch 1.13.1
 sentence-transformers>=2.1.0,<=2.2.2; python_version<'3.11'
-transformers[torch]>=4.12.0,<=4.27.4; python_version<'3.11'
+transformers[torch]>=4.31.0,<=4.33.2; python_version<'3.11'
```
pquentin, line 25: A newer version of transformers is needed to get access token support. This was already upgraded in setup.py, but not here. I'll have to avoid the duplication in the future.

eland/ml/pytorch/transformers.py
```diff
@@ -584,7 +584,7 @@ def __init__(
         self,
         *,
         model_id: str,
-        access_token: Optional[str],
+        access_token: Optional[str] = None,
```
pquentin, line 587: The order of `access_token` does not matter because those are kwargs-only parameter that can't be passed conditionally.

joshdevins: Stylistically, I always prefer default args are at the end, but I don't know if this is more or less Pythonic. Either way is fine, just curious

joshdevins (approved):
LGTM. I see two changes in this PR though. All of the other changes, aside from the argument default are related to something else? It's fine in one PR, not a big deal, just wanted to understand why the other changes.

pquentin:
Thanks. My intention is always to make a single change per pull request. In that case, there is indeed a single fix in the core Eland package. The rest of the pull request is to fix the tests that should have caught the bug but were skipped due to a technicality in pytest. Does that make sense?

eland/ml/pytorch/transformers.py
pquentin: Ah, that's an interesting consideration. I don't know if there's one choice that's more Pythonic than the other (I haven't seen enough such cases to tell).

I don't have a strong preference so I am happy to change this.

pquentin:
Tested the latest commit with `eland_import_hub_model` and the `PyTorchModel` API, merging.

PR #623, opened 2023-10-18
[NLP] Tests for NLP model configurations 

davidkyle:
The configurations generated for the models uploaded to Elasticsearch are derived from transformer models config, this test asserts on various settings in the generated configurations. 

A new function is added for finding the model's `max_sequence_length` as the value is found in different places. Specifically the previous method of looking up the model in the tokenizer's `max_model_input_sizes` map sometimes returns `None`

```
getattr(self._tokenizer, "max_model_input_sizes", dict()).get(self._model_id)
```

If the above returns `None` then the tokenizer's `model_max_length` property is checked and if that is `None` the model configuration is checked.

If `max_sequence_length` cannot be found an exception is thrown. This changes the behaviour as previously it was left unset and Elasticsearch would pick a default value of 512, which is good because the model gets uploaded but bad because the error is silently ignored and may result in an incorrect setting. 

One suggestion is to allow the user to manually specify a `max_sequence_length` in the case where it cannot be found.


eland/ml/pytorch/nlp_ml_model.py
```diff
@@ -317,11 +317,9 @@ def __init__(
         input: TrainedModelInput = TrainedModelInput(field_names=["text_field"]),
         metadata: t.Optional[dict] = None,
         model_type: t.Union["t.Literal['pytorch']", str] = "pytorch",
-        default_field_map: t.Optional[t.Mapping[str, str]] = None,
```
davidkyle, line 320: Unused so I removed this

tests/ml/pytorch/test_transformer_pytorch_model_pytest.py
```diff
@@ -67,9 +60,6 @@
         ES_VERSION < (8, 0, 0),
         reason="This test requires at least Elasticsearch version 8.0.0",
     ),
-    pytest.mark.skipif(
```
davidkyle, line 70: sklearn is not required for this test

eland/ml/pytorch/transformers.py
```diff
@@ -756,7 +775,7 @@ def _create_config(
             ),
         )
 
-    def _create_traceable_model(self) -> TraceableModel:
+    def _create_traceable_model(self) -> Tuple[Any, TraceableModel]:
```
joshdevins, line 778: I find this change introduces some complexity that could maybe be encapsulated better by just adding the `transformer_model` to `_TransformerTraceableModel`. We do this with the tokenizer, which gets saved as a property of that class. It would make it cleaner to do that with the underlying HF model, and would mean we don't need to return tuples here. If it doesn't make sense to return a `TraceableModel` then maybe it's worth switching to return `_TransformerTraceableModel`.

joshdevins (changes requested):
Generally LGTM, just a small change requested on the `_create_traceable_model`.

joshdevins (approved):

PR #625, opened 2023-10-19
[NLP] Support E5 small multi-lingual

### LINKED ISSUE #583: Model multilingual-e5-small fails to start
At present, the model is processed and uploaded fine, but when starting the model, it fails:

```
forward() is missing value for argument 'token_type_ids'.
```

The model is trained from Multilingual-MiniLM which is a BERT model, but uses the XLM-RoBERTa tokenizer. Since we wrap models based on their architecture and not on the tokenizer type, the BERT model expects input that isn't coming from the XLM-RoBERTa tokenizer. We should consider changing how we decide which wrapper to use (three inputs or two) based on the tokenizer instead.

Note that the base and large model variants work fine because they are XLM-RoBERTa models, and use the corresponding tokenizer.

davidkyle:
E5 small multi lingual is based on a BERT architecture which made the code that traces the model assume the model takes 4 parameters to its `forward()` function. However, the model uses the XLMRoBERTa tokenizer which produces 2 inputs, this lead to an error when evaluating the model as it complained of missing arguments.

The fix is to use the tokenizer type to determine the number of inputs to the model's `forward()` function 

Closes #583


joshdevins (approved):
I know we've done this for a while, but is there a way to reflect on the tokenizer itself to find out the outputs it produces, instead of just having a hard-coded list of tokenizers that are 2 or 4 outputs?

joshdevins:
Does this close #583 or is another PR pending?

davidkyle:
> Does this close https://github.com/elastic/eland/issues/583 or is another PR pending?

One more PR queued up to add some testing but we can consider #583 closed after this

PR #627, opened 2023-11-05
Fix docs build

pquentin:
Some dependencies like numpy are pinned to versions that do not support Python 3.12. Python 3.10 is the latest version supported by Eland.


davidkyle (approved):
LGTM

PR #630, opened 2023-11-08
Make demo notebook runnable in Colab

leemthompo:
### Summary:

- Upload dataset directly in notebook
- Create index with mappings
- Ignore docs that fail to index
- Group imports and installs

### Related work: 

This is a more simplistic approach to https://github.com/elastic/eland/pull/604

## [Visual diff](https://github.com/elastic/eland/assets/32779855/59863755-d06f-4732-afeb-d6c9079c68e3)
(Just the opening section — no changes to the actual examples)


pquentin (approved):
Thanks! LGTM.

Great work here on understanding what was needed to run on Colab and not doing more than necessary while still producing a great result.


I'm seeing ids like `lswztIsBt5-t_OatRvrU` in places like `lswztIsBt5-t_OatRvrU` instead of monotically increasing ids, and I'd like to know why, but it's not blocking IMO.

leemthompo:
Should be good now using 

```python
def generate_actions(data):
    for idx, entry in enumerate(data, start=0):  # Use monotonically increasing IDs starting from 0
        yield {
            "_index": FLIGHTS_INDEX_NAME,
            "_id": idx,  # Use the current index as the document ID
            "_source": entry
        }```

pquentin (approved):
Thanks! Much better. It's indeed how the tests do it (well the use Panda's itterrows, but that's the same).

PR #632, opened 2023-11-15
Fix colab link

leemthompo:
The Colab button doesn't work in readthedocs, so just using standard markdown link format.


pquentin:
@leemthompo Since the "Open in Colab" image is so recognizable, I restored it using the Markdown image syntax + the link syntax. What do you think? https://eland--632.org.readthedocs.build/en/632/examples/demo_notebook.html

leemthompo:
@pquentin Perfect LGTM! 🪄


Unrelated, but I found a weirdness where `latest` doesn't map to v8.11.x on site.

https://eland.readthedocs.io/en/v8.11.0/examples/demo_notebook.html

versus

https://eland.readthedocs.io/en/latest/examples/demo_notebook.html

pquentin:
> Unrelated, but I found a weirdness where `latest` doesn't map to v8.11.x on site.

`latest` include the latest commits to main. Your notebook fix landed just after the 8.11.0 release, which is why it isn't there. Does that make sense?

leemthompo:
@pquentin I can't approve my own PR 😄

pquentin (approved):

pquentin:
Hmm indeed my approval was enough. Interesting. Sorry about the noise!

PR #636, opened 2023-11-21
Bumpy numpy and shap

pquentin:
Shap is maintained again and supports newer numpy version.


davidkyle (approved):
LGTM

Great news that Shap is being maintained and updated

PR #642, opened 2023-12-08
[NLP] Add prefix_string config option to the import model hub script

davidkyle:
The prefix_strings option was added in https://github.com/elastic/elasticsearch/pull/102089 to support the E5 model.

The new params to `eland_import_hub_model` are `--search-prefix` and `--ingest-prefix`

### Example usage
```
eland_import_hub_model       
  --url 'http://localhost:9200'
  --hub-model-id intfloat/multilingual-e5-small       
  --task-type text_embedding 
  --search-prefix "query: " 
  --ingest-prefix "passage: "
```

This option is only available in Elasticsearch 8.12 and depends on https://github.com/elastic/elasticsearch-specification/pull/2363

### DIFF
--- a/eland/cli/eland_import_hub_model.py
+++ b/eland/cli/eland_import_hub_model.py
@@ -128,6 +128,19 @@ def get_arg_parser():
         "--ca-certs", required=False, default=DEFAULT, help="Path to CA bundle"
     )
 
+    parser.add_argument(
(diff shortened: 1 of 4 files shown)


joshdevins (approved):
Is this documented somewhere in docs too?

PR #645, opened 2024-01-03
Accept LTR inference config when creating model

demjened:
This PR adds support for importing a model with a specific `inference_config`. This allows us to use regression models with the `learning_to_rank` inference type that come with externally supplied configuration.

Example:
```python
MLModel.import_model(
    es_client,
    model_id,
    regressor,
    feature_names,
    es_if_exists="replace",
    es_compress_model_definition=compress_model_definition,
    inference_config={
        "learning_to_rank": {
            "feature_extractors": [
                {
                    "query_extractor": {
                        "feature_name": "title_bm25",
                        "query": {"match": {"title": "{{query_string}}"}},
                    }
                },
                {
                    "query_extractor": {
                        "feature_name": "imdb_rating",
                        "query": {
                            "script_score": {
                                "query": {"exists": {"field": "imdbRating"}},
                                "script": {"source": 'return doc["imdbRating"].value;'},
                            }
                        },
                    }
                },
            ]
        }
    },
)
```

The change is backward compatible - if `inference_config` is not passed, it will be set to the default `{<model_type>: {}}`.

### DIFF
--- a/eland/ml/ml_model.py
+++ b/eland/ml/ml_model.py
@@ -23,7 +23,7 @@ import numpy as np
 from eland.common import ensure_es_client, es_version
 from eland.utils import deprecated_api
 
-from .common import TYPE_CLASSIFICATION, TYPE_REGRESSION
+from .common import TYPE_CLASSIFICATION, TYPE_LEARNING_TO_RANK, TYPE_REGRESSION
 from .transformers import get_model_transformer
 
 if TYPE_CHECKING:
@@ -130,6 +130,11 @@ class MLModel:
         >>> # Delete model from Elasticsearch
         >>> es_model.delete_model()
         """
+        if self.model_type not in (TYPE_CLASSIFICATION, TYPE_REGRESSION):
+            raise NotImplementedError(
+                f"Prediction for type {self.model_type} is not supported."
+            )
+
         docs: List[Mapping[str, Any]] = []
         if isinstance(X, np.ndarray):
 
@@ -215,7 +220,9 @@ class MLModel:
         inference_config = self._trained_model_config["inference_config"]
         if "classification" in inference_config:
             return TYPE_CLASSIFICATION
-        elif "regression" in inference_config or "learning_to_rank" in inference_config:
+        elif "learning_to_rank" in inference_config:
+            return TYPE_LEARNING_TO_RANK
+        elif "regression" in inference_config:
             return TYPE_REGRESSION
         raise ValueError("Unable to determine 'model_type' for MLModel")
 
--- a/eland/ml/ml_model.py
+++ b/eland/ml/ml_model.py
@@ -174,6 +174,7 @@ class MLModel:
                         "inference": {
                             "model_id": self._model_id,
                             "inference_config": {self.model_type: {}},
+                            # "inference_config": self.inference_config,
                             field_map_name: {},
                         }
                     }
@@ -215,7 +216,7 @@ class MLModel:
         inference_config = self._trained_model_config["inference_config"]
         if "classification" in inference_config:
             return TYPE_CLASSIFICATION
-        elif "regression" in inference_config:
+        elif "regression" in inference_config or "learning_to_rank" in inference_config:
             return TYPE_REGRESSION
         raise ValueError("Unable to determine 'model_type' for MLModel")
 
@@ -254,7 +255,7 @@ class MLModel:
         classification_weights: Optional[List[float]] = None,
         es_if_exists: Optional[str] = None,
         es_compress_model_definition: bool = True,
-        inference_config: Optional[Dict[str, Dict[str, Any]]] = None
+        inference_config: Optional[Dict[str, Dict[str, Any]]] = None,
     ) -> "MLModel":
         """
         Transform and serialize a trained 3rd party model into Elasticsearch.
@@ -372,7 +373,7 @@ class MLModel:
         )
         serializer = transformer.transform()
         model_type = transformer.model_type
-        default_inference_config = {model_type: {}}
+        default_inference_config: Dict[str, Any] = {model_type: {}}
 
         if es_if_exists is None:
             es_if_exists = "fail"
--- a/eland/ml/ml_model.py
+++ b/eland/ml/ml_model.py
@@ -261,7 +261,7 @@ class MLModel:
         classification_weights: Optional[List[float]] = None,
         es_if_exists: Optional[str] = None,
         es_compress_model_definition: bool = True,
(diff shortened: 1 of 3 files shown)


tests/ml/test_ml_model_pytest.py
```diff
@@ -306,6 +306,68 @@ def test_decision_tree_regressor(self, compress_model_definition):
...
+        regressor.fit(training_data[0], training_data[1])
+
+        # Serialise the models to Elasticsearch
+        feature_names = [f"feature_{i}" for i in range(2)]
+        model_id = "test_learning_to_rank"
+        inference_config = {
+            "learning_to_rank": {
+                "feature_extractors": [
+                    {
+                        "query_extractor": {
```
afoucret, line 324: Those query extractors should be suitable to be used with the movie index:

```python
feature_extractors = [
  {
    "query_extractor": {
      "feature_name": "title_bm25",
      "query": {"match": { "title": "{{query_string}}" } }
    }
  },
  {
    "query_extractor": {
       "feature_name": "imbd_rating",
       "query": {
         "script_score": {
           "query": {"exists": { "field": "imdbRating" }},
           "script": {"source": 'return doc["imdbRating"].value;'},
       }
     }
  }
]

# Feature names also need to be updated!
feature_names = [extractor['query_extractor']['feature_name']  for extractor in feature_extractors]
```

tests/ml/test_ml_model_pytest.py
```diff
@@ -306,6 +306,68 @@ def test_decision_tree_regressor(self, compress_model_definition):
...
+        saved_inference_config = response.body["trained_model_configs"][0][
+            "inference_config"
+        ]
+        assert "learning_to_rank" in saved_inference_config
+        saved_ltr_config = saved_inference_config["learning_to_rank"]
+        assert all(
+            item in saved_ltr_config.items()
+            for item in inference_config["learning_to_rank"].items()
+        )
+
```
afoucret, line 367: You should be able to use the model as a learning to rank rescorer on the movie index:

```
ES_TEST_CLIENT.search(
  index = MOVIES_INDEX_NAME,
  query = { 
    "multi_match": { 
      "fields": ["title", "actors", "directors" , "plot"], 
      "query": "planet of the apes" 
    }
  },
  rescore = { 
    "learning_to_rank": { 
      "model_id": model_id, 
      "params": {"query_string": "planet of the apes" } 
    } 
  }
)
```

eland/ml/ml_model.py
```diff
@@ -254,6 +261,7 @@ def import_model(
         classification_weights: Optional[List[float]] = None,
         es_if_exists: Optional[str] = None,
         es_compress_model_definition: bool = True,
+        inference_config: Optional[Dict[str, Dict[str, Any]]] = None,
```
afoucret, line 264: This is more a semantic change but I suspect we should use `Mapping` instead of `Dict` here.

eland/ml/ml_model.py
```diff
@@ -367,6 +379,7 @@ def import_model(
         )
         serializer = transformer.transform()
         model_type = transformer.model_type
+        default_inference_config: Dict[str, Any] = {model_type: {}}
```
afoucret, line 382: Change to `Mapping[str, Mapping[str, Any]]` to stay consistent with inference_config.

tests/ml/test_ml_model_pytest.py
afoucret: Potentially, the scores extracted by the feature extractors are too close and our completely randomly trained model returns the same score for them.

I did try a completely different test on my env:

```python
# Execute search with rescoring and verify document order
search_response = ES_TEST_CLIENT.search(
    index=MOVIES_INDEX_NAME,
    query={"terms": {"_id": ["tt1318514", "tt0071562"] } },
    rescore={
        "learning_to_rank": {
            "model_id": model_id,
            "params": {"query_string": "planet of the apes"},
        }
    },
)

doc_scores = [hit['_score'] for hit in search_response['hits']['hits']]

# Assert that:
# - all documents from the query are present
# - all documents have been rescored (score != 1.0)
# - document scores are unique
assert len(search_response['hits']['hits']) == 2
assert all(score != float(1) for score in doc_scores)
assert all(doc_scores.count(score) == 1 for score in doc_scores)
```

I rescore two documents that are very different so it is unlikely that they get close scores during feature extraction.

Also, it seems to me a little bit trappy to assume our random regression model will rescrore to preserve an order (it could do the inverse). Instead of doc ordering, I prefer to test that all docs have been rescored and that all scores are different.

WDYT?

demjened: I support the idea because testing the rescorer's behavior is out of scope for the unit test. So as long as we verify it does something to the scores, we're good.
I'll make the code changes as suggested.

tests/ml/test_ml_model_pytest.py
```diff
@@ -306,6 +319,98 @@ def test_decision_tree_regressor(self, compress_model_definition):
...
+        # Assert that:
+        # - all documents from the query are present
+        # - all documents have been rescored (score != 1.0)
+        # - document scores are unique
+        doc_scores = [hit["_score"] for hit in search_result["hits"]["hits"]]
+        assert len(search_result["hits"]["hits"]) == 2
+        assert all(score != float(1) for score in doc_scores)
+        assert all(doc_scores.count(score) == 1 for score in doc_scores)
```
demjened, line 403: @afoucret Looks like document scores are not unique. Among the two consequent runs of this test (`compress_model_definition=True|False`) I tend to get equal scores for the two documents in the hit list:
```
doc_scores [147.05865, 109.10927] # run 1
doc_scores [139.4883, 139.4883] # run 2 - this assertion fails 
```

I don't understand where the randomness comes from; shouldn't the results be deterministic with a static test fixture and LTR configuration?

afoucret: The randomness comes from the model which is trained from random data. Let me see if I can find a way to workaround this.

afoucret: For your PR, it seems impossible to control the randomness in the model. I think we should remove the tests that check the score are all different for now:
assert all(doc_scores.count(score) == 1 for score in doc_scores)
When I will add the feature logger, I will be able to

demjened: Ok, removed.

afoucret (approved):
Everything looks fine to me.
I would like someone from the Eland team to chime in.

Also, I will replace the inference_config with a LearnToRankModelConfig object in a follow up PR next week.

davidkyle (approved):
LGTM

PR #648, opened 2024-01-05
LTR feature logger

afoucret:
The PR provides tools for LTR query feature extraction into Eland.

1. Some base class have been defined to declare the LTR config:

```python
ltr_config = LTRModelConfig(
  feature_extractors=[
    QueryFeatureExtractor(
      feature_name="title_bm25", 
      query={"match": {"title": "{{query}}"}},
    ),
    QueryFeatureExtractor(
      feature_name="popularity",
      query={
        "script_score": {
          "query": {"exists": {"field": "popularity"}},
          "script": {"source": "return doc['popularity'].value"},
        }
      },
    )
  ]
)
```

2. A feature logger can be created using:

```python
feature_logger = FeatureLogger(es_client="http://localhost:9200",  es_index="my-index", ltr_model_config=ltr_model_config)
```

3. Document features can be extracted for a set of query params using:

```python
feature_logger.extract_features(
  query_params={ "query": "my search query" },
  doc_ids= ["doc-1", "doc-2"] 
)
```


eland/ml/ltr/ltr_model_config.py
```diff
@@ -0,0 +1,153 @@
...
+                    for feature_extractor in self.feature_extractors
+                ]
+            }
+        }
+
+    def feature_names(self) -> List[str]:
+        """
+        List of the feature names for the model.
+        """
+        if not hasattr(self, "_feature_names"):
```
ioanatia, line 137: we seem to reuse this patern in multiple methods:

```
if not hasattr(self, attr_name)
  # assign attribute
return attribute
```
this seems like a good candidate for using the `cached_property` decorator: https://docs.python.org/3/library/functools.html#functools.cached_property

I see from the README that Eland supports python 3.8, 3.9, 3.10. So `cached_property` should be available.

https://github.com/elastic/eland/blob/926f0b9b5c5980e42db9e6e119a290a2eb4c7a60/README.md?plain=1#L57

ioanatia (commented):
Since `movies.json.gz` is only used for tests, is it possible to reduce its size? with `7.43 MB` it seems to be the largest artefact in `tests`.

eland/ml/ltr/ltr_model_config.py
afoucret: I did change the implementation to use cached_property. Thank you for the tip.

ioanatia (approved):

NOTICE.txt
```diff
@@ -50,3 +50,6 @@ Permission is hereby granted, free of charge, to any person obtaining a copy of
 The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
 
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+--
+This product contains a adapted version of the "us-national-parks" dataset, https://data.world/kevinnayar/us-national-parks, by Kevin Nayar, https://data.world/kevinnayar, is licensed under CC BY, https://creativecommons.org/licenses/by/4.0/legalcode
```
kderusso, line 55: Typo - this should be "`an` adapted version"

PR #649, opened 2024-01-05
Add XGBRanker and transformer

demjened:
This PR adds [XGBRanker](https://xgboost.readthedocs.io/en/stable/python/python_api.html#xgboost.XGBRanker) to the list of supported model types.


eland/ml/transformers/xgboost.py
```diff
@@ -203,6 +209,9 @@ def is_objective_supported(self) -> bool:
             "reg:pseudohubererror",
             "reg:linear",
             "reg:logistic",
+            "rank:pairwise",
```
afoucret, line 212: Just a minor comment but it seems that we could be more specific on the supported objective with something like:

if isinstance(XGBRanker, model):
  return objective in [rank:pairwise, rank:map, rank:ndcg]
 
return objective in [reg:logistic, ....]

demjened: I updated the logic. Note for this to work I had to store the model as a new instance variable:
```
self._regressor_model = model
```

(We cannot use `self._model` as it's always a [Booster](https://github.com/elastic/eland/pull/649/files#diff-2c78ef83042f3d461f76e152e17e7008946667a6ca910b79ae8c2dd99565b825R199).)

afoucret: Thank you for iterating on this. 
Cause we are touching a very specific area of the code, I think it is OK to merge.

PR #651, opened 2024-01-16
LTR improvements

afoucret:
This PR contains two improvements for the LTR feature in Eland.

- Feature logger now extract `NaN` when a query does not match a query. It is consistent with what is done in Elasticsearch.

- A specific method (`import_ltr_model`) has been added to import LTR model with support for a `ltr_model_config` object of class `LTRModelConfig`. The `inference_model` params is removed from the `import_model` method

**Example:**

```python
ed.ml.MLModel.import_ltr_model(
  es_client=es_client,
  model=ranker,
  model_id='ltr-model-xgb',
  ltr_model_config=ltr_model_config,
  es_if_exists = 'replace'
)
```

The previous version was more difficult to use (and could lead to inconsistencies):

```python
ed.ml.MLModel.import_model(
  es_client=es_client,
  model=ranker,
  model_id='ltr-model-xgb',
  feature_names=ltr_model_config.feature_names,
  inference_config=ltr_model_config.to_dict(),
  es_if_exists = 'replace'
)
```

### DIFF
--- a/eland/ml/ml_model.py
+++ b/eland/ml/ml_model.py
@@ -24,6 +24,7 @@ from eland.common import ensure_es_client, es_version
 from eland.utils import deprecated_api
 
 from .common import TYPE_CLASSIFICATION, TYPE_LEARNING_TO_RANK, TYPE_REGRESSION
+from .ltr import LTRModelConfig
 from .transformers import get_model_transformer
 
 if TYPE_CHECKING:
@@ -266,7 +267,6 @@ class MLModel:
         classification_weights: Optional[List[float]] = None,
         es_if_exists: Optional[str] = None,
         es_compress_model_definition: bool = True,
-        inference_config: Optional[Mapping[str, Mapping[str, Any]]] = None,
     ) -> "MLModel":
         """
         Transform and serialize a trained 3rd party model into Elasticsearch.
@@ -342,10 +342,6 @@ class MLModel:
             JSON instead of raw JSON to reduce the amount of data sent
             over the wire in HTTP requests. Defaults to 'True'.
 
-        inference_config: Mapping[str, Mapping[str, Any]]
-            Model inference configuration. Must contain a top-level property whose name is the same as the inference
-            task type.
-
         Examples
         --------
         >>> from sklearn import datasets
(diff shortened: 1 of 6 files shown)


eland/ml/ml_model.py
```diff
@@ -380,6 +376,120 @@ def import_model(
...
+            model_id=model_id,
+            model=model,
+            feature_names=feature_names,
+            classification_labels=classification_labels,
+            classification_weights=classification_weights,
+            es_if_exists=es_if_exists,
+        )
+
+    @classmethod
+    def import_ltr_model(
```
kderusso, line 391: Question, can this be used to import models for any use cases besides LTR?

kderusso (approved):
Overall LGTM a couple of non-blocking questions though

eland/ml/ml_model.py
```diff
@@ -380,6 +376,120 @@ def import_model(
...
+        cls,
+        es_client: Union[str, List[str], Tuple[str, ...], "Elasticsearch"],
+        model_id: str,
+        model: Union[
+            "DecisionTreeRegressor",
+            "RandomForestRegressor",
+            "XGBRanker",
```
demjened, line 398: Are `XGBRanker` (and possibly other model types) meant to be imported only with this function? If so, consider removing the type from `import_model()`'s `model` property, since that function doesn't support passing in an LTR `inference_config` anymore.

eland/ml/ml_model.py
```diff
@@ -380,6 +376,120 @@ def import_model(
...
+            - xgboost.XGBRegressor
+                - only the following objectives are supported:
+                    - "reg:squarederror"
+                    - "reg:linear"
+                    - "reg:squaredlogerror"
+                    - "reg:logistic"
+                    - "reg:pseudohubererror"
+
+        ltr_model_config: LTRModelConfig
+            The LTR model configuration is used to configure features extractors for the LTR model.
```
demjened, line 441:
```suggestion
            The LTR model configuration is used to configure feature extractors for the LTR model.
```

eland/ml/ml_model.py
```diff
@@ -380,6 +376,120 @@ def import_model(
...
+            - elasticsearch-py parameters or
+            - elasticsearch-py instance
+
+        model_id: str
+            The unique identifier of the trained inference model in Elasticsearch.
+
+        model: An instance of a supported python model. We support the following model types for LTR prediction:
+            - sklearn.tree.DecisionTreeRegressor
+            - sklearn.ensemble.RandomForestRegressor
+            - xgboost.XGBClassifier
```
demjened, line 422: - `XGBClassifier` is not in the list of [supported model types](https://github.com/elastic/eland/pull/651/files#diff-ec1a6cd34e674e32e217dbce70918329e9853078e06e6c5fe0bff3d5498a3554R395-R401)
- `LGBMRegressor` is in the list of supported model types but it's not listed here

eland/ml/ml_model.py
```diff
@@ -380,6 +376,120 @@ def import_model(
...
+            "XGBRegressor",
+            "LGBMRegressor",
+        ],
+        ltr_model_config: LTRModelConfig,
+        es_if_exists: Optional[str] = None,
+        es_compress_model_definition: bool = True,
+    ) -> "MLModel":
+        """
+        Transform and serialize a trained 3rd party model into Elasticsearch.
+        This model can then be used as an learning_to_rank rescorer in the Elastic Stack.
```
demjened, line 408:
```suggestion
        This model can then be used as a learning_to_rank rescorer in the Elastic Stack.
```

demjened (approved):
LGTM, minor comments and questions

eland/ml/ml_model.py
afoucret: Nope. For other use case you have to use the general purpose `import_model` method.

eland/ml/ml_model.py
afoucret: Good catch. I did remove XGBClassifier from supported models in the doc and added LGBMRegressor.

eland/ml/ml_model.py
afoucret: We can imagine one that want to use a ranker model in a non-LTR context, so I would recommend to keep it supported in the regular import_model method.

afoucret:
@elasticmachine run elasticsearch-ci/docs

pquentin:
buildkite test this please

PR #652, opened 2024-01-17
Run Buildkite docs jobs in pull requests from forks

pquentin:
This will avoid having to run `buildkite test this please` in all pull requests created from forks by members of the elastic organization.


pquentin:
buildkite test this please

PR #654, opened 2024-01-23
Fix missing value support for XGBRanker.

afoucret:
When exporting XGBoost trees to ES, it was assumed are always using the left branch of the tree. 
In fact it is false and it is needed to configure it using a conditional on `default_left`.

Tests have been updated to check predicted score in _search match the one from the ranker model. It should allows us to catch any problem like this in the future.

### DIFF
--- a/eland/ml/transformers/xgboost.py
+++ b/eland/ml/transformers/xgboost.py
@@ -107,6 +107,7 @@ class XGBoostForestTransformer(ModelTransformer):
                 decision_type=self._node_decision_type,
                 left_child=self.extract_node_id(row["Yes"], curr_tree),
                 right_child=self.extract_node_id(row["No"], curr_tree),
+                default_left=row["Yes"] == row["Missing"],
                 threshold=float(row["Split"]),
                 split_feature=self.get_feature_id(row["Feature"]),
             )
--- a/tests/ml/test_ml_model_pytest.py
+++ b/tests/ml/test_ml_model_pytest.py
@@ -23,7 +23,7 @@ import pytest
 
 import eland as ed
 from eland.ml import MLModel
-from eland.ml.ltr import LTRModelConfig, QueryFeatureExtractor
+from eland.ml.ltr import FeatureLogger, LTRModelConfig, QueryFeatureExtractor
 from tests import (
     ES_TEST_CLIENT,
     ES_VERSION,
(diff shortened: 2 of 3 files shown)


tests/ml/test_ml_model_pytest.py
```diff
@@ -321,13 +321,27 @@ def test_decision_tree_regressor(self, compress_model_definition):
...
+        rng = np.random.default_rng()
+        qid = rng.integers(0, 3, size=X.shape[0])
+
+        # Sort the inputs based on query index
+        sorted_idx = np.argsort(qid)
+        X = X[sorted_idx, :]
+        y = y[sorted_idx]
+        qid = qid[sorted_idx]
+
+        ranker = XGBRanker(objective=objective)
```
demjened, line 343: With this change (switching to a ranker from a regressor) this test is now almost the same as `test_xgb_ranker()`, except it has a few additional assertions. Can we merge them or is there a reason for keeping both?

eland/ml/transformers/xgboost.py
```diff
@@ -107,6 +107,7 @@ def build_tree_node(self, row: pd.Series, curr_tree: int) -> TreeNode:
                 decision_type=self._node_decision_type,
                 left_child=self.extract_node_id(row["Yes"], curr_tree),
                 right_child=self.extract_node_id(row["No"], curr_tree),
+                default_left=row["Yes"] == row["Missing"],
```
demjened, line 110: I don't know how the tree works internally. What does this line mean - we're traversing the left subtree if the "Yes" value is the same as the "Missing" value? Could you summarize what that means?

demjened (approved):
LGTM with a few questions.

eland/ml/transformers/xgboost.py
afoucret: Each node of the tree (except leafs that own value) have 2 children nodes: left and right.
Based on the value of the feature we decide to go left or right. If the feature is missing we can decide (default_left) to go to the left or to the right (training decide for us).

When exporting xgboost the tree is represented as a table and:
- `row['Yes']` contains the id of the left node
- `row['No']` contains the id of the right node
- `row['Missing']` contains the id of the node when a value is missing and is always equals to `row['Yes']` or `row['No']`

So this code convert this xgboost specific structure into one that ES understand.

tests/ml/test_ml_model_pytest.py
afoucret: `test_xgb_ranker` is doing testing for using a XGBRanker as a regressor. It is a test that I would like to keep since it allows to track regression in the inference process without worrying about others aspects from LTR (feature extractions, rescoring, ...)

In contrast `test_learning_to_rank` test the whole feature.

eland/ml/transformers/xgboost.py
demjened: Gotcha, thanks for the clarification!

PR #661, opened 2024-02-07
Implement eland.DataFrame.to_json

bartbroere:
Dumping an Elastic index was the right solution for me at first. After a while csv did not offer all the guarantees I was looking for, since csv records for example can span multiple lines if any of the values contain line breaks.

For that reason, [JSON lines](https://jsonlines.org/) was a more suitable format. Pandas' `DataFrame.to_json` can generate ".jsonl" files by suppling `lines=True, orient='records'`. This also lets us reuse the earlier solution that streams output from Elastic and appends it to a file, eliminating the need to be able to fit the entire dataset in memory.

This pull request implements streaming an Elasticsearch index to `.jsonl`, while falling back to running `to_pandas().to_json(...)` if streaming is a bit harder to do.


eland/operations.py
```diff
@@ -1250,6 +1250,40 @@ def to_csv(  # type: ignore
...
+                    ):
+                        print(f"{datetime.now()}: read {processed} rows")
+                    output = df.to_json(
+                        orient=orient,
+                        lines=lines,
+                        **kwargs,
+                    )
+                    w.write(output.strip())
+                    w.write("\n")
+            return "".join(result) or None
```
pquentin, line 1281: What is the point of this? result is always the empty array

eland/operations.py
```diff
@@ -1250,6 +1250,40 @@ def to_csv(  # type: ignore
...
+    def to_json(  # type: ignore
+        self,
+        query_compiler: "QueryCompiler",
+        show_progress=False,
+        path_or_buf=None,
+        orient=None,
+        lines=False,
+        **kwargs,
+    ):
+        if orient == "records" and lines is True and isinstance(path_or_buf, str):
```
pquentin, line 1262: Any reason not to handle path objects and file-like objects?

pquentin (commented):
Thanks for this!

eland/operations.py
bartbroere: Good point, this will contain something now.

eland/operations.py
bartbroere: With the latest commit it should handle file-like objects and paths as strings. Let me check on how pandas handles the other kinds of path and implement that as well.

I'll be adding some tests too.

In an earlier commit this was a lot easier to handle, but I mistakenly assumed the Pandas 2.0.0 signature of `to_json`. In that version the `mode` parameter was added (similar as `to_csv` provides. Pandas 1.5 unfortunately does not have this yet, complicating things a bit.

pquentin (approved):
Thanks! LGTM.

pquentin:
buildkite test this please

tests/dataframe/test_to_json_pytest.py
```diff
@@ -0,0 +1,142 @@
...
+            ),
+            pandas.read_json(
+                root_dir / "dataframe" / "results" / "pathlib_pandas_to_json.jsonl",
+                lines=True,
+                orient="records",
+            ),
+        )
+
+    def test_to_json_with_other_buffer(self):
+        root_dir = Path(ROOT_DIR)
```
pquentin, line 99: lint fails because this isnt' used

eland/operations.py
```diff
@@ -1250,6 +1251,47 @@ def to_csv(  # type: ignore
...
+        self,
+        query_compiler: "QueryCompiler",
+        path_or_buf=None,
+        orient=None,
+        lines=False,
+        **kwargs,
+    ):
+        if orient == "records" and lines is True:
+            result = []
+            processed = 0
```
pquentin, line 1264: lint fails because this isn't used

bartbroere:
```suggestion
```

Thanks! I decided last-minute to leave out the `show_progress` parameter in this method, but that of course makes this integer no longer required.

tests/dataframe/test_to_json_pytest.py
bartbroere:
```suggestion
```

Thanks, removing it!

bartbroere:
> Thanks! LGTM.

Nice! I removed the two linting errors. Sorry about that.

pquentin:
buildkite test this please

pquentin:
buildkite test this please

PR #665, opened 2024-02-15
Add feedback request to README

pquentin:
We're adding this to all language clients. Is it a good fit for Eland too?


ezimuel (approved):
I think this is a good idea. I would also create an issue with this feedback request

pquentin:
buildkite test this please

PR #671, opened 2024-02-22
Upgrade torch to 2.1.2

### LINKED ISSUE #669: Upgrade PyTorch to version 2.1.2 and
Pytorch was upgraded to 2.1.2 in https://github.com/elastic/ml-cpp/pull/2588 can now be upgraded to the same version.

Models traced in Pytorch 2.1.2 may not be compatible with version 1.13.1 of libtorch used in Elasticsearch 8.12 and earlier. This was the case with an earlier upgrade to Pytorch 1.11  after which #481  was implemented to prevent Eland installing a model into an earlier version of Elasticsearch

davidkyle:
Elasticsearch is upgraded to PyTorch v2.1.2 for v8.13.0, this change makes Eland use the same version

Due to potential backwards compatibility issues with models traced in PyTorch 2.1.2 running on earlier versions the `eland_import_hub_model` script will not upload models to a cluster that is earlier than Elasticsearch v8.13.0

The tests caught a problem tracing the [valhalla/distilbart-mnli-12-6](https://huggingface.co/valhalla/distilbart-mnli-12-6) model:

```
ValueError: Attention using SDPA can not be traced with torch.jit.trace when no attention_mask is provided. To solve this issue, please either load your model with the argument `attn_implementation="eager"` or pass an attention_mask input when tracing the model. 
```

The error comes from the decoder layer, in the transformers code the decoder attention mask is not initialised and appears to be added in [this PR](https://github.com/huggingface/transformers/pull/26572). 

> or pass an attention_mask input when tracing the model. 

The error message is misleading as the input included `attention_mask`. 

The change was added in transformers v4.36.0 released in December 2023. Rather than work around the problem by injecting the missing parameters the simplest solution is to pin the transformers version to 4.35.2.

Closes #669

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -31,6 +31,6 @@ steps:
           - '3.9'
           - '3.8'
         stack:
-          - '8.13-SNAPSHOT'
-          - '8.12'
+          - '8.13.0-SNAPSHOT'
+          - '8.12.2'
     command: ./.buildkite/run-tests
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -31,6 +31,6 @@ steps:
           - '3.9'
           - '3.8'
         stack:
-          - '8.11-SNAPSHOT'
-          - '8.12-SNAPSHOT'
+          - '8.13-SNAPSHOT'
+          - '8.12'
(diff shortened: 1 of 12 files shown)


valeriy42 (approved):
LGTM 👍

Issue #675, opened 2024-03-19
Learning to rank tests failing


davidkyle:
Several learning to rank tests are failing with the error:
```
Arrays are not almost equal to 2 decimals
```

```
FAILED tests/ml/test_ml_model_pytest.py::TestMLModel::test_learning_to_rank[rank:ndcg-False]
FAILED tests/ml/test_ml_model_pytest.py::TestMLModel::test_learning_to_rank[rank:map-True]
FAILED tests/ml/test_ml_model_pytest.py::TestMLModel::test_learning_to_rank[rank:map-False]
FAILED tests/ml/test_ml_model_pytest.py::TestMLModel::test_learning_to_rank[rank:pairwise-True]
FAILED tests/ml/test_ml_model_pytest.py::TestMLModel::test_learning_to_rank[rank:pairwise-False]
```

The first failure reproduces on my system with this command
```
pytest tests/ml/test_ml_model_pytest.py::TestMLModel::test_learning_to_rank[rank:ndcg-False]
```

davidkyle:
Closed by #685

PR #676, opened 2024-03-19
Mute the Learning to Rank tests

davidkyle:
For #675


pquentin (approved):
Thanks! LGTM. Maybe pinning xgboost would help us here too: https://github.com/elastic/eland/pull/660

PR #678, opened 2024-03-21
Update overview.asciidoc to replace tuple reference to API Key

iuliaferoli:
Same fix as in https://github.com/elastic/elasticsearch-py/pull/2477 for the one reference to the API Key tuple in the Eland documentation


pquentin (approved):
Thanks! LGTM.

PR #681, opened 2024-03-26
Add Python 3.11 to support matrix

davidkyle:
Python 3.11 support was blocked by the version of PyTorch used. Now that PyTorch has been upgraded 3.12 can be added to the support matrix

PyTorch 2.2.0 supports Python 3.12 so we will have to wait until the next PyTorch upgrade to enable Python 3.12


.buildkite/pipeline.yml
```diff
@@ -27,9 +27,9 @@ steps:
     matrix:
       setup:
         python:
+          - '3.11'
           - '3.10'
           - '3.9'
-          - '3.8'
```
pquentin, line 32: Are you dropping Python 3.8 support on purpose? If yes we need to update `python_requires` in setup.py.

pquentin (commented):
We need to update `CLASSIFIERS` in setup.py. And I would suggest running `pyupgrade --py38-plus **/*.py` or `pyupgrade --py39-plus **/*.py` depending on the actually mininum supported version.

The scientific ecosystem has mostly dropped Python 3.8, so I would be fine dropping it here too: https://scientific-python.org/specs/spec-0000/

.buildkite/pipeline.yml
davidkyle: I thought Python 3.8 could be removed from CI but as it is still supported that would be a silly thing to do. I've reverted this change

davidkyle:
`CLASSIFIERS` is updated and the result of running `pyupgrade --py38-plus **/*.py` is in commit 04c38289d8cfe0b7f0c0fac34564424a64319f46

pquentin (approved):
Thanks! LGTM.

davidkyle:
Thanks @pquentin

PR #684, opened 2024-03-28
Add from kql

bartbroere:


bartbroere:
I'll be mostly offline during April, so this is just a heads up that this PR will not see much progress for a few weeks. I didn't forget about it though, and plan to continue working on it in May.

PR #667, opened 2024-02-15
Document TLS/SSL options for import script

pquentin:
Closes #663

PR #688, opened 2024-05-02
Support HTTP proxies in eland_import_hub_model

pquentin:
Depends on #667 

Tested by running `python -m dummyserver.proxy` using tag 2.1.0 of urllib3 (I broke the command-line proxy in later commits of urllib3) and using `export HTTPS_PROXY=http://localhost:8888/` to upload a model with Eland. Without the proxy running, the upload fails.

### DIFF
--- a/docs/guide/machine-learning.asciidoc
+++ b/docs/guide/machine-learning.asciidoc
@@ -154,6 +154,20 @@ underscores `__`.
 
 --
 
+[discrete]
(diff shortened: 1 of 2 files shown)


davidkyle (approved):
LGTM 

I tested with `proxy.py`. It might be worth mentioning that if security is disabled use the HTTP_PROXY env var

miguelgrinberg (approved):

PR #691, opened 2024-05-06
Stop duplicating requirements

pquentin:
Today we specify `elasticsearch>=8.3,<9` in four places. In setup.py, as expected for the package metadata to be correct, but also:

 * in requirements.txt, which was unused, so I removed it
 * in requirements-dev.txt, which can simply refer to `.[extras]` to install all packages with its extras
 * in noxfile.py, where installing `.` is enough to make lint pass (also we don't need pre-releases anymore)
 * in docs/requirements-docs.txt which was out of date anyway

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ extras = {
     "scikit-learn": ["scikit-learn>=1.3,<1.4"],
     "lightgbm": ["lightgbm>=2,<4"],
     "pytorch": [
+        "requests<3",
         "torch==2.1.2",
         "tqdm",
         "sentence-transformers>=2.1.0,<=2.3.1",
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,7 @@ extras = {
     "lightgbm": ["lightgbm>=2,<4"],
     "pytorch": [
         "torch==2.1.2",
+        "tqdm",
         "sentence-transformers>=2.1.0,<=2.3.1",
         "transformers[torch]>=4.31.0,<4.36.0",
     ],
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,5 +7,5 @@ build:
 
 python:
   install:
-    - requirements: docs/requirements-docs.txt
     - path: .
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -60,6 +60,7 @@
     "lightgbm": ["lightgbm>=2,<4"],
     "pytorch": [
         "torch==2.1.2",
+        "tqdm",
```
pquentin, line 64: tqdm is needed in PytorchModel, so let's declare it explicitly even though transformers requires it already.

setup.py
```diff
@@ -59,7 +59,9 @@
     "scikit-learn": ["scikit-learn>=1.3,<1.4"],
     "lightgbm": ["lightgbm>=2,<4"],
     "pytorch": [
+        "requests<3",
```
pquentin, line 62: requests is required for proxy support in eland_import_hub_model, so let's declare it explicitly even though transformers requires it already.

davidkyle (commented):
I always have difficulties building and installing LightGBM for vanilla Python3 on macOS, that was one benefit of having the old `extras` as I did not need to install LightGBM for my purposes.

I tested this in clean venv, running `python -m pip install -r requirements-dev.txt ` fails because LightGBM requires wheel, `python -m pip install wheel` fixed that but then the LightGBM build failed because `libomp` is missing.

I _think_ Conda ships with libomp so it is not a problem there. 

Is there a way to have default all but excluded certain parts if I don't want them?

pquentin:
Thanks for looking. I don't understand how that is relevant to that pull request though. On main requirements-dev.txt also contains lightgbm. If you don't want it, it is easier to edit it out with this pull request since it's only in setup.py.

davidkyle:
I now realised I can install Eland with the PyTorch dependencies but without LightGBM by running

```
python -m pip install .[pytorch]
```

That, essentially what I was asking for and its existed for a very long time but I didn't realise so I had a hacky workaround involving requirements.txt.

Thanks for cleaning up the install config.

davidkyle (approved):
LGTM

PR #695, opened 2024-05-21
Fix missing method str.removeprefix in Python 3.8

bartbroere:
Eland still supports Python 3.8, so str.removeprefix can be called and raise an `AttributeError`.

Once Eland drops Python 3.8 support this if/else can be removed (only keeping the code in the if).


pquentin (changes requested):
Thanks! I don't think this is worth a condition, I would use the slice and mention this can be replaced by removeprefix when we remove Python 3.8.

bartbroere:
> I would use the slice and mention this can be replaced by removeprefix when we remove Python 3.8.

That sounds a bit less verbose indeed. I committed the suggestion.

pquentin (approved):

pquentin:
buildkite test this please

pquentin:
buildkite test this please

pquentin:
buildkite test this please

PR #698, opened 2024-06-06
Make model upload compatible with serverless clusters

### LINKED ISSUE #696: Eland should be Serverless aware when throwing incompatible ES version errors
Currently, eland checks the ES version whek checking compatible.

```
2024-05-31 12:44:50,385 INFO : Establishing connection to Elasticsearch
2024-05-31 12:44:50,808 INFO : Connected to cluster named '<Serverless Project ID>' (version: 8.11.0)
2024-05-31 12:44:50,808 ERROR : Eland uses PyTorch version 2.1.2 which is incompatible with Elasticsearch versions prior to 8.13. Please upgrade Elasticsearch to at least version 8.13
```
On Serverless, this is problematic because Serverless does not have the notion of version but it will always return an arbitrary 8.11.0 version number (e.g. in GET /) for backwards compatibility reasons (it does not mean that automatic upgrades on Serverless is not working).  So eland will need to be updated to check against `build_flavor` instead of `version.number`. Otherwise, Serverless ES projects will always fail with model loading:

```
"build_flavor": "serverless"
```

davidkyle:
As reported in #696 serverless clusters always return version `8.11.0` which makes the model upload script think that it does not support the upgraded version of PyTorch. The fix is to check for  `build_flavor == serverless`

Closes #696

### DIFF
--- a/eland/cli/eland_import_hub_model.py
+++ b/eland/cli/eland_import_hub_model.py
@@ -32,6 +32,7 @@ import textwrap
 from elastic_transport.client_utils import DEFAULT
 from elasticsearch import AuthenticationException, Elasticsearch
 
+from eland._version import __version__
 from eland.common import parse_es_version
 
 MODEL_HUB_URL = "https://huggingface.co"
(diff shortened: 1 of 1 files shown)


eland/cli/eland_import_hub_model.py
```diff
@@ -195,6 +196,17 @@ def get_es_client(cli_args, logger):
 
 def check_cluster_version(es_client, logger):
     es_info = es_client.info()
+
+    if (
+        "build_flavor" in es_info["version"]
+        and es_info["version"]["build_flavor"] == "serverless"
+    ):
```
pquentin, line 203: It's OK to assume `build_flavor` is set, it was here in the 7.x days too: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/targz.html#_checking_that_elasticsearch_is_running

```suggestion
    if es_info["version"]["build_flavor"] == "serverless":
```

pquentin (approved):
Thanks! LGTM. It works well.

eland/cli/eland_import_hub_model.py
davidkyle: Oh it failed running against a locally built and run Elasticsearch maybe it's not in dev builds. I'd prefer to keep this to be safe

PR #693, opened 2024-05-14
Fix fields missing from the result hits (fixes #692)

bartbroere:
This is a hotfix that solves the issues in our data warehouse. Some more work is needed to reproduce this issue with one of the test datasets. Also see the analysis in the related issue #692 

To do:
- [x] Regenerate notebook tests

Fixes #692

Issue #692, opened 2024-05-14
Fields (multi-fields?) specified in eland.Dataframe columns variable are not always returned


bartbroere:
I encountered the issue that a field specified in `eland.Dataframe`'s columns parameter ended up as an empty array in the Dataframe (all NaNs), even though it definitely was not empty.

After setting some breakpoints in `eland.operations` I found a fix that works on our index.
I did not yet reproduce it with the `flights` or `ecommerce` sample data yet, but I could work on that later.

bartbroere:
I think this problem might occur with multi-fields specifically: https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html

bartbroere:
I think we could reproduce this issue if we add a `copy_to` to the testing indices: https://www.elastic.co/guide/en/elasticsearch/reference/current/copy-to.html

I'm trying that now.

pquentin:
Thanks. Would you mind pasting here a JSON response from Elasticsearch where _source is a list and fields is set?

PR #707, opened 2024-06-27
Deal with the possibility of lists in multi-fields

bartbroere:
While running the new release, it looks like my previous pull request #693 didn't deal with an edge case while checking whether a field was present. It is possible for a field to be a list of dicts as well.

On top of that it feels a bit like painting myself into a corner fixing the missing fields #692. The example indices eland uses for testing aren't as complex (and nested) as the mappings I'm dealing with in our internal data warehouse.

### DIFF
--- a/eland/operations.py
+++ b/eland/operations.py
@@ -1543,14 +1543,22 @@ def quantile_to_percentile(quantile: Union[int, float]) -> float:
     return float(min(100, max(0, quantile * 100)))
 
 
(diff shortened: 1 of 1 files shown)


pquentin:
buildkite test this please

pquentin (approved):
Thanks! LGTM.

pquentin:
The complexity is unfortunate but we don't have much choice here.

PR #708, opened 2024-07-02
Remove input fields from exported LTR models

afoucret:
## Context:

In Elasticsearch >= 8.15.0, we update the `PUT _ml/trained_models/<model_id>` API for LTR models and specifying `input.field_names` is not supported anymore. 

Previously, the field was required and eland was putting feature names into it as a hack.

The Elasticsearch PR: https://github.com/elastic/elasticsearch/pull/109437

## PR content:

The behavior now depends of the es version and the `input.field_names` is set only for ES < 8.15.0

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -32,6 +32,7 @@ steps:
           - '3.9'
           - '3.8'
         stack:
-          - '8.13.0-SNAPSHOT'
+          - '8.15.0-SNAPSHOT'
+          - '8.14.1'
           - '8.12.2'
     command: ./.buildkite/run-tests
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -34,5 +34,4 @@ steps:
         stack:
           - '8.15.0-SNAPSHOT'
           - '8.14.1'
-          - '8.12.2'
     command: ./.buildkite/run-tests
--- a/eland/ml/ml_model.py
+++ b/eland/ml/ml_model.py
@@ -533,20 +533,20 @@ class MLModel:
             ),
         )
 
-        put_trained_model_kwargs = {
-            "model_id": model_id,
-            "inference_config": inference_config or default_inference_config,
-            "input": trained_model_input,
-        }
-
         if es_compress_model_definition:
-            put_trained_model_kwargs["compressed_definition"] = (
-                serializer.serialize_and_compress_model()
+            ml_model._client.ml.put_trained_model(
+                model_id=model_id,
+                inference_config=inference_config or default_inference_config,
+                input=trained_model_input,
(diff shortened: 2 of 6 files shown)


afoucret:
@elasticmachine rerun buildkite/eland

pquentin (approved):
Thanks, LGTM!

Note that we ask users to use an Eland version that matches the minor version of the target Elasticsearch cluster, so this check is not strictly necessary, but since it's easy enough to do, it makes sense to add. Indeed, this recommendation is not enforced today.

.buildkite/pipeline.yml
```diff
@@ -32,6 +32,7 @@ steps:
           - '3.9'
           - '3.8'
         stack:
-          - '8.13.0-SNAPSHOT'
+          - '8.15.0-SNAPSHOT'
+          - '8.14.1'
           - '8.12.2'
```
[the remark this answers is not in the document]
afoucret: Done

eland/ml/ml_model.py
```diff
@@ -523,18 +525,23 @@ def _import_model(
         elif es_if_exists == "replace":
             ml_model.delete_model()
 
+        trained_model_input = None
+        is_ltr = next(iter(inference_config)) is TYPE_LEARNING_TO_RANK
+        if not is_ltr or es_version(es_client) < (8, 15):
```
davidkyle, line 530: You should also check `or build_flavor == serverless` as serverless reports its version as 8.11. 

See #698 for an example

davidkyle:
I'd really like to get rid of the required `input.field_names` for uploading HuggingFace models too, your ES PR contains has a clever solution I might extend for the HF NLP models

eland/ml/ml_model.py
afoucret: This is a good catch. I will add this.

tests/ml/test_ml_model_pytest.py
```diff
@@ -438,6 +448,7 @@ def test_learning_to_rank(self, objective, compress_model_definition):
             pass
 
         # Clean up
+        ES_TEST_CLIENT.cluster.health(index='.ml-*', wait_for_active_shards='all')
```
afoucret, line 451: I did add this to prevent test flakiness.

Indeed, the the deletion of the model sometimes failed because of a search_phase exception. Under the hood it is caused by the cleaning of the stats happening during deletion that is occurring before the ml-stats index has all his shards assigned.

This is probably something to be fixed in ES (and clearly not related to this PR).

eland/ml/ml_model.py
afoucret: @davidkyle I introduced a new method to check is the build flavor is serverless in commons and updated the code.
I have the intuition that it will be used in other places in the future.

I also updated the existing check in `eland/cli/eland_import_hub_model.py` to use this new method.

PR #713, opened 2024-07-25
Default truncation to `second` for text similarity

davidkyle:
NLP models have 3 truncation settings: `FIRST`, `SECOND` and `NONE`

`FIRST` means truncate the first input. In most cases there is only 1 input (e.g for text embeddings) so this is a sensible default.
`SECOND` means truncate the second input. Task types with 2 inputs are extractive question answering where the question is one input and the context the other. Text Similarity takes has 2 inputs.
`NONE` means don't truncate and window the input.

For text similarity the first input is usually the shorter input, for example it might be the query text in a rerank operation. In this situation it is better to truncate the second input. This change makes that the default.

### DIFF
--- a/eland/ml/pytorch/transformers.py
+++ b/eland/ml/pytorch/transformers.py
@@ -770,6 +770,10 @@ class TransformerModel:
             tokenization_config.span = 128
             tokenization_config.truncate = "none"
 
+        if self._task_type == "text_similarity":
+            print("is text sime")
+            tokenization_config.truncate = "second"
+
         if self._traceable_model.classification_labels():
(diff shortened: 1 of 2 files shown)


eland/ml/pytorch/transformers.py
```diff
@@ -770,6 +770,10 @@ def _create_config(
             tokenization_config.span = 128
             tokenization_config.truncate = "none"
 
+        if self._task_type == "text_similarity":
+            print("is text sime")
```
pquentin, line 774: leftover print?

```suggestion
```

pquentin (approved):
Thanks! LGTM. Just need to remove the extra print.

PR #718, opened 2024-08-15
Upgrade PyTorch to version 2.3.1

davidkyle:
Elasticsearch has upgraded to PyTorch to 2.3.1 in 8.16 and 8.15.2. This PR updates Eland to use the same version.

The usual compatibility issues apply, Eland will refuse to upload to an Elasticsearch cluster that has is using an earlier version of PyTorch. The range of compatible versions of Sentence Transformers and HuggingFace Transformers has been updated to incorporate the latest release of the same major version. 

Elasticsearch is upgraded in https://github.com/elastic/ml-cpp/pull/2705

### DIFF
--- a/eland/cli/eland_import_hub_model.py
+++ b/eland/cli/eland_import_hub_model.py
@@ -229,8 +229,6 @@ def check_cluster_version(es_client, logger):
 
     sem_ver = parse_es_version(es_info["version"]["number"])
     major_version = sem_ver[0]
-    minor_version = sem_ver[1]
-    patch_version = sem_ver[2]
 
     # NLP models added in 8
     if major_version < 8:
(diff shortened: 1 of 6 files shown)


eland/cli/eland_import_hub_model.py
```diff
@@ -238,13 +239,13 @@ def check_cluster_version(es_client, logger):
         )
         exit(1)
 
-    # PyTorch was upgraded to version 2.1.2 in 8.13
+    # PyTorch was upgraded to version 2.3.1 in 8.15.1
     # and is incompatible with earlier versions
-    if major_version == 8 and minor_version < 13:
+    if major_version == 8 and minor_version < 15 and patch_version < 1:
```
pquentin, line 244: If the user is on 8.15.0, this will translate to: `if True and False and True` and they won't get the error. I would suggest using `sem_ver` directly instead:

```suggestion
    if sem_ver < (8, 15, 1):
```

eland/cli/eland_import_hub_model.py
```diff
@@ -230,6 +230,7 @@ def check_cluster_version(es_client, logger):
     sem_ver = parse_es_version(es_info["version"]["number"])
     major_version = sem_ver[0]
     minor_version = sem_ver[1]
+    patch_version = sem_ver[2]
```
pquentin, line 233: Those two lines will be unnecessary if we adopt my suggestion below.

pquentin (approved):
Thanks! LGTM, but we need to make sure this really makes it into 8.15.1 before merging.

eland/cli/eland_import_hub_model.py
davidkyle: Thanks. I've made this change and updated the version check to 8.15.2

pquentin (approved):
Thanks! LGTM.

PR #722, opened 2024-09-08
Restore public Dockerfile

pquentin:
Since non-latest Wolfi images are not public, we allow users to continue building with Dockerfile being Debian-based.


davidkyle (approved):
LGTM

PR #724, opened 2024-09-26
Set valid timestamp for all rows

pquentin:
The JSON version of the flights dataset had timestamp fields with only dates, no times. This was preventing Kibana file upload to detect them correctly, see https://github.com/elastic/kibana/pull/193744. I used `00:00:00` here as already done in the dataframe version.

(I also removed commented out code.)


pquentin:
Superseded by #742.

PR #726, opened 2024-09-27
Fix path in docker upload example

davidkyle:


pquentin (commented):
Should we fix the example above, line 132? It still is `--hub-model-id /PATH/TO/MODEL` but should be `--hub-model-id /PATH/TO/MODELS`.

davidkyle:
> Should we fix the example above, line 132? It still is --hub-model-id /PATH/TO/MODEL but should be --hub-model-id /PATH/TO/MODELS.

Both examples use model singular, this make sense to me

PR #728, opened 2024-10-01
Fix Docker image build on Linux

pquentin:
I tested the Docker image buid on macOS, which does not need a version. And I failed to test a staging build + push on Buildkite before releasing, sorry.

This will require a new release, Eland 8.15.2.


pquentin:
Since building the ARM image takes one hour on Buildkite, that I don't want to introduce macOS workers and that the macOS image is the least likely to break as it's simpler, I'm only building the Linux images now. Merging, thanks for the review!

PR #732, opened 2024-10-07
Allow reading Elasticsearch certs in Wolfi image

pquentin:
The config/certs directory of Elasticsearch is not readable by other users and groups. This work in the public image, which uses the root user, but the Wolfi image does not. Using the same user id fixes the problem.

Thanks @legacyworld for the report.


ezimuel (approved):
LGTM

PR #734, opened 2024-10-11
Revert "Allow reading Elasticsearch certs in Wolfi image"

pquentin:
Reverts elastic/eland#732. Setting a different user actually breaks model upload, as Hugging Face writes to a cache folder unconditionally (even with caching disabled: https://huggingface.co/docs/datasets/v2.17.1/en/package_reference/main_classes#datasets.disable_caching). In my tests with `--ca-certs` I only checked that we could read the cert, unfortunately.


ezimuel (approved):
LGTM

PR #738, opened 2024-11-07
Add deprecation warning for ESGradientBoostingModel subclasses

valeriy42:
Introduce a warning indicating that exporting data frame analytics models as ESGradientBoostingModel subclasses is deprecated and will be removed in version 9.0.0.

The implementation of ESGradientBoostingModel relies on importing undocumented private classes that were changed in 1.4 to https://github.com/scikit-learn/scikit-learn/pull/26278. This dependency makes the code difficult to maintain, while the functionality is not widely used by users. Therefore, we will deprecate this functionality in 8.16 and remove it completely in 9.0.0.

### DIFF
--- a/eland/__init__.py
+++ b/eland/__init__.py
@@ -15,6 +15,8 @@
 #  specific language governing permissions and limitations
 #  under the License.
 
+import warnings
+
 from ._version import (  # noqa: F401
     __author__,
     __author_email__,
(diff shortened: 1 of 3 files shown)


pquentin:
I have made a few changes to make sure this warning is seen by users:

 * Switch from `FutureDeprecationWarning` to `DeprecationWarning`. The former is ignored by default since the deprecation is only intended to happen in the future.
 * Create a subclass `ElandDeprecationWarning` in order to call `warnings.simplefilter("default", category=ElandDeprecationWarning)` which will always show the Eland warnings, without having to use `python -Wall`. (We don't want to show *all* deprecation warnings, only the Eland ones.)
 * Fix the stacklevel to be 2 instead of 3 (to show the line that instantiates the model)
 * Add a test for this warning

I also tested manually:

```python
from eland.ml.exporters.es_gb_models import ESGradientBoostingModel

model = ESGradientBoostingModel(..., ...)
print(model)
```

This correctly identifies the third line as the culprit:

```
$ python t.py
/.../eland/t.py:3: ElandDeprecationWarning: Exporting data frame analytics
models as ESGradientBoostingModel subclasses is deprecated and will be removed in version 9.0.0.
  model = ESGradientBoostingModel(1, 2)
<eland.ml.exporters.es_gb_models.ESGradientBoostingModel object at 0x102e359d0>
```

Please take a look.

pquentin (approved):

Issue #737, opened 2024-11-01
max-model-input-length argument not working in prepare input, lead to python kernel crash


JasonRichard:
eland/eland/ml/pytorch/transformers.py

max-model-input-length argument not working in prepare input,
max_length uses max_length from model weight/config.json
leads to python kernel crash when max_length is too large

```python
class _TraceableTextEmbeddingModel(_TransformerTraceableModel):
    def _prepare_inputs(self) -> transformers.BatchEncoding:
        return self._tokenizer(
            "This is an example sentence.",
            padding="max_length", #may lead to python kernel crash
            return_tensors="pt",
        )
```
![Image](https://github.com/user-attachments/assets/72faa1aa-ac00-4284-87f5-08183b21ae1b)

maybe change it to 'longest' ?

davidkyle:
Hi @JasonRichard 

Which model were you importing when you saw the error and please can you share the command you do use so I can reproduce the error. And which version of Eland are you using?

JasonRichard:
baai/bge-m3 with default max_model_input_length = 8192
with a 32 GB RAM server

upload using max_model_input_length = 1024
crashes when reaching 
```python
    def sample_output(self) -> Tensor:
        inputs = self._compatible_inputs()
>>>>    return self._model(*inputs)
```
in file eland/ml/pytorch/transformers.py
inputs is well prepared but the calculation exceed the resources

quent1fvr:
Hello, I face the exact same issue as described by @JasonRichard 
Did you find any solution @JasonRichard ? 
Any idea on how to fix it @davidkyle 

Thanks in advance

davidkyle:
> maybe change it to 'longest' ?

This did not help I got the same crash when using `longest`. 

What did help was upgrading the transformers library after which I was able to upload the model to Elasticsearch and use it. You can see what I did here: https://github.com/elastic/eland/pull/752. That PR is draft I will make the change proper for the next release

davidkyle:
Closed by https://github.com/elastic/eland/pull/752

PR #752, opened 2025-02-06
Fix tracing baai/bge-m3

davidkyle:
Fixes a kernel crash importing https://huggingface.co/BAAI/bge-m3 as described in #737

The required change is upgrading the version of transformers.

```
eland_import_hub_model \
  --url 'http://localhost:9200' \
  -u elastic-admin -p elastic-password \
  --hub-model-id 'baai/bge-m3' \
  --task-type text_embedding --insecure --max-model-input-length=8192
```


pquentin (approved):
Thanks! LGTM.

PR #768, opened 2025-04-03
Update the documentation to reflect the partial support of eland/scki…

redcinelli:
Update the documentation to reflect the partial support of eland/sckittlearn


leemthompo (approved):
lgtm

PR #770, opened 2025-04-11
Update ES stack versions in build matrix

davidkyle:
Use the latest 8.18.0 snapshot version in CI


pquentin:
Sorry, I missed this pull request. What we currently do in `main` (9.x releases) is to only test against 9.x versions (currently only 9.0, but we should likely add 9.1.0-SNAPSHOT):

https://github.com/elastic/eland/blob/87380ef71605ef6be858d23ee916cdce83252c73/.buildkite/pipeline.yml#L47-L48


However the 8.x branch is indeed outdated, so I'll update the target branch.

PR #776, opened 2025-04-22
Add 9.1.0 snapshot to build matrix

davidkyle:
And use 9.0.0 rather than the snapshot build

PR #777, opened 2025-04-22
Add 9.1.0 Snapshot to Build and Fix test_ml_model Tests to Normalized Expected Scores if Min Score is Less Than Zero

markjhoy:
Adds the 9.1.0-SNAPSHOT to the build.

Also fixes a bug in the tests:

The LTR rescoring in Elasticsearch 9.1+ has the ability to normalize the scores for LTR if any score is negative (as Lucene cannot handle negative scores). This updates the model tests to ensure scores are normalized as well.

The normalization process essentially adds the negative score to all scores, sliding the scoring window upwards, so that all scores are >= 0.0

Should fix: https://github.com/elastic/eland/pull/776


tests/ml/test_ml_model_pytest.py
```diff
@@ -330,6 +330,17 @@ def test_learning_to_rank(self, objective, compress_model_definition):
             ],
             reverse=True,
         )
+
+        if (ES_VERSION[0] == 8 and ES_VERSION >= (8, 19)) or (ES_VERSION >= (9, 1)):
```
afoucret, line 334: if (ES_VERSION[0] == 8 and ES_VERSION >= (8, 19)) or (ES_VERSION >= (9, 1)):

Should also apply to serverless.
A cluster feature would have been a great option for this.

afoucret (approved):

PR #780, opened 2025-04-23
Forbid Elasticsearch 8 client or server

pquentin:
Today, Eland 9.0 works fine on an Elasticsearch 8.15.2 server, but since we're no longer testing for it, should we disable it?


davidkyle (commented):
Please also remove the version check from this test

https://github.com/elastic/eland/blob/main/tests/ml/pytorch/test_pytorch_model_upload_pytest.py#L41

davidkyle:
> Please also remove the version check from this test

I opened #792 to make do this

davidkyle (approved):
> Eland 9.0 works fine on an Elasticsearch 8.15.2 server, but since we're no longer testing for it, should we disable it?

Yes to disabling. Let's keep things simple and require a 9 Eland to talk to a 9 cluster.

PR #783, opened 2025-04-23
Add 8.19.0 Snapshot to Build and Fix test_ml_model Tests to Normalized Expected Scores if Min Score is Less Than Zero

markjhoy:
Backport to `8.x` from #777 

Adds the 8.19.0-SNAPSHOT to the build.

Fixes a bug in the tests:

The LTR rescoring in Elasticsearch 8.19+ has the ability to normalize the scores for LTR if any score is negative (as Lucene cannot handle negative scores). This updates the model tests to ensure scores are normalized as well.

The normalization process essentially adds the negative score to all scores, sliding the scoring window upwards, so that all scores are >= 0.0

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -48,4 +48,5 @@ steps:
           - '8.18.0'
           - '8.16.6'
           - '8.14.1'
+          - '8.19.0-SNAPSHOT'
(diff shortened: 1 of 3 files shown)


markjhoy:
buildkite test this

PR #785, opened 2025-04-24
Upgrade PyTorch to 2.5.1

davidkyle:
PyTorch was upgraded to 2.5.1 in ml-cpp on the 8.18 and 9.0 branches in https://github.com/elastic/ml-cpp/pull/2800

The backport of this will require this version check to be updated https://github.com/elastic/eland/blob/main/eland/cli/eland_import_hub_model.py#L242


pquentin (approved):
Thanks! LGTM. Is it expected that `eland_import_hub_model` currently works on 9.0 without this change?

davidkyle:
> Is it expected that eland_import_hub_model currently works on 9.0 without this change?

Yes, PyTorch is backwards compatible. If the Elasticsearch cluster is running a newer version of PyTorch than the Eland client the model uploaded by Eland is still compatible.

PR #787, opened 2025-04-24
Python 3.13

pquentin:


davidkyle:
The 3.13 builds failed with - `Building wheel for sentencepiece (pyproject.toml) did not run successfully.`

[sentencepiece](https://pypi.org/project/sentencepiece/) is required for the NLP model tokenizers

pquentin:
Correct, this is the current blocker! If it does not get resolved by the time, say, Python 3.14 is out, we can consider workarounds like allowing Python 3.13 for Pandas but not NLP models, or making sentencepiece an optional dependency.

The workaround suggested by the SentencePiece is to use conda, which isn't really practical. Closing for now.

PR #790, opened 2025-04-24
Add 9.0.1 check for LTR rescoring for tests

markjhoy:
Follow on to https://github.com/elastic/eland/pull/777

Now that the LTR rescoring bounded window fix will be in 9.0.1 [via this PR](https://github.com/elastic/elasticsearch/pull/127345), this adds 9.0.1+ to the condition to check to see if the test needs to rescore the expected scores.


davidkyle (approved):
LGTM

PR #792, opened 2025-04-30
Remove redundant version checks in tests

davidkyle:
The main branch does not run CI against any 8.x clusters so there is reason to check the 8.x version.


pquentin (approved):
Thanks, LGTM!

